PR #39, opened 2026-01-08
Allow NIOHTTPServer to use NIOAsyncTestingChannel for tests

aryan-25:
Motivation:

Currently, we perform end-to-end tests by starting `NIOHTTPServer` on localhost and setting up a HTTP client to send requests / observe responses. `NIOHTTPServer` internally uses [`ServerBootstrap`](https://swiftpackageindex.com/apple/swift-nio/2.92.1/documentation/nioposix/serverbootstrap) to set up the server, but `ServerBootstrap` does not expose the child channels it creates for each connection, making `NIOHTTPServer` difficult to test closely.

We can improve this by refactoring `NIOHTTPServer` to also (in tests) allow a [`NIOAsyncTestingChannel`](https://github.com/apple/swift-nio/blob/main/Sources/NIOEmbedded/AsyncTestingChannel.swift#L87) to be used as the channel the server runs on, and manually feed/extract data from it to test different expectations.

Modifications:

- Moved the `serveInsecureHTTP1_1` and `serveSecureUpgrade` methods to new files (`NIOHTTPServer+HTTP1_1.swift` and `NIOHTTPServer+SecureUpgrade.swift`).
  - Split the server channel and child channel setup into different methods.
  - Added `serveInsecureHTTP1_1WithTestChannel` and `serveSecureUpgradeWithTestChannel` methods, which set up the input `NIOAsyncTestingChannel` as the channel the server runs on.
- Added `HTTP1ClientServerProvider` and `HTTPSecureUpgradeClientServerProvider`, which can be used in tests to send requests / observe responses. These types currently have the following methods (this is very much open for discussion):
   - `static func withProvider(handler:body:)`:
      - Sets up the server with the provided request handler closure, and vends a provider instance to the `body` closure.
    - `func withConnectedClient(body:)`:
      - Sets up a connection and vends the client `NIOAsyncChannel` to the `body` closure which can be used to send requests / observe responses in terms of Swift HTTP types.
      - For Secure Upgrade, `withConnectedClient` vends a `NegotiatedConnection` enum, which, for HTTP/1.1 gives you the client `NIOAsyncChannel`, or for HTTP/2, a `HTTP2StreamManager` which has an `openStream()` method returning a client stream channel.
- Added two example test-cases in `NIOHTTPServerEndToEndTests.swift` which use these new provider types.

Result:

Infrastructure in place to write end-to-end tests.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -35,6 +35,10 @@ let package = Package(
             targets: ["HTTPServer"]
         )
     ],
+    traits: [
+        .trait(name: "SwiftConfiguration"),
+        .default(enabledTraits: ["SwiftConfiguration"]),
+    ],
     dependencies: [
         .package(
             url: "https://github.com/FranzBusch/swift-collections.git",
@@ -48,6 +52,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.36.0"),
         .package(url: "https://github.com/apple/swift-nio-extras.git", from: "1.30.0"),
         .package(url: "https://github.com/apple/swift-nio-http2.git", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-configuration", from: "1.0.0"),
     ],
     targets: [
         .executableTarget(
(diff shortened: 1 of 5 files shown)


Sources/HTTPServer/NIOHTTPServer+HTTP1_1.swift
```diff
@@ -0,0 +1,109 @@
...
+                            channel: http1Channel,
+                            handler: handler
+                        )
+                    }
+                }
+            }
+        }
+    }
+
+    func serveInsecureHTTP1_1WithTestChannel(
```
gjcairo: Can we move this to some utility in the tests module?

PR #44, opened 2026-01-09
Respect maximumCount argument in request reader

aryan-25:
Motivation:

Currently, we do not use the `maximumCount` argument in `HTTPRequestConcludingAsyncReader`'s `read(maximumCount:body:)` method.

Modifications:

Updated `HTTPRequestConcludingAsyncReader/read(maximumCount:body:)` to throw a `LimitExceeded` error when the bytes available exceed `maximumCount`.

Result:

The `maximumCount` argument is now respected in `HTTPRequestConcludingAsyncReader/read(maximumCount:body:)`.

### DIFF
--- a/Sources/NIOHTTPServer/HTTPRequestConcludingAsyncReader.swift
+++ b/Sources/NIOHTTPServer/HTTPRequestConcludingAsyncReader.swift
@@ -50,15 +50,12 @@ public struct HTTPRequestConcludingAsyncReader: ConcludingAsyncReader, ~Copyable
                 case finished
 
                 enum ReadingBodyState {
-                    // Not yet received any request body parts
-                    case initial
+                    // All received bytes have been consumed; no excess bytes need to be stored.
+                    case noExcess
 
                     // `read` was called with a `maximumCount` value that was lower than the bytes available. The excess
                     // bytes are stored here so they can be dispensed in future calls to `read`.
                     case excess(ByteBuffer)
-
-                    // No excess bytes currently needing to be stored
-                    case noExcess
                 }
             }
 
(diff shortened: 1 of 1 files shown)


Sources/HTTPServer/LimitExceeded.swift
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Data available in a read buffer exceeded the caller-specified maximum count.
+struct LimitExceeded: Error {
```
gjcairo: I'm not sure we need this based on the other comment but if we do, I think it makes more sense to have a namespace for this, either by nesting it inside the `HTTPRequestConcludingAsyncReader` or by having some other error namespace and nesting it there.

Sources/NIOHTTPServer/HTTPRequestConcludingAsyncReader.swift
```diff
@@ -41,8 +41,93 @@ public struct HTTPRequestConcludingAsyncReader: ConcludingAsyncReader, ~Copyable
...
+            enum State {
+                // The request body is currently being read: expecting more request body parts or a request end part.
+                case readingBody(ReadingBodyState)
+
+                // The request end part was received. We have finished.
+                case finished
+
+                enum ReadingBodyState {
+                    // Not yet received any request body parts
+                    case initial
```
gjcairo: I would personally get rid of this `initial` case. I see `ReadingBodyState` as more of a "`BufferedDataAvailability`" or something like that (not particularly suggesting the name change): there either is some data buffered that we need, or there isn't. We're not really handling `initial` and `noExcess` differently, so let's simplify the state machine where we can.

gjcairo (approved):

PR #48, opened 2026-01-14
Fix formatting

gjcairo:
There are several formatting issues that need to be fixed:
- YAML formatting
- Missing/incorrect license headers
- `forEach` formatting rule needs to be disabled in a couple of places


gjcairo:
The formatting check will fail because only the nightly formatter understands `nonisolated(nonsending)` and we're using 6.2. For now we'll have to live with it failing.

PR #52, opened 2026-01-27
Split NIOHTTPServer into separate module

aryan-25:
Motivation:

Both the abstract HTTP server API and the concrete NIO implementation are currently contained in the same module. We should separate the concrete NIO implementation into its own module.

Modifications:

Added a new `NIOHTTPServer` target and updated import statements. The resulting structure is:

```
Sources/HTTPServer
├── HTTPRequestContext.swift
├── HTTPResponseSender.swift
├── HTTPServer.swift
├── HTTPServerClosureRequestHandler.swift
├── HTTPServerRequestHandler.swift
└── RequestResponseMiddlewareBox.swift
Sources/NIOHTTPServer
├── HTTPRequestConcludingAsyncReader.swift
├── HTTPResponseConcludingAsyncWriter.swift
├── NIOHTTPServer.swift
├── NIOHTTPServer+ConnectionContext.swift
├── NIOHTTPServer+HTTP1_1.swift
├── NIOHTTPServer+ListeningAddress.swift
├── NIOHTTPServer+SecureUpgrade.swift
├── NIOHTTPServer+SwiftConfiguration.swift
├── NIOHTTPServerConfiguration.swift
├── NIOSSL+X509.swift
└── SocketAddress.swift
```

Result:

The abstract HTTP server API and the concrete NIO implementation are now split in different modules.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -77,12 +77,7 @@ let package = Package(
             name: "HTTPServer",
             dependencies: [
                 "AsyncStreaming",
-                .product(name: "DequeModule", package: "swift-collections"),
-                .product(name: "BasicContainers", package: "swift-collections"),
                 .product(name: "HTTPTypes", package: "swift-http-types"),
-                .product(name: "Logging", package: "swift-log"),
-                .product(name: "NIOCore", package: "swift-nio"),
-                .product(name: "NIOHTTP1", package: "swift-nio"),
             ],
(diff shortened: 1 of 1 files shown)


gjcairo (approved):
LGTM. 
Note: API break check fails but that's expected as we're splitting the only module in two.

PR #54, opened 2026-02-06
Handle request stream ending before receiving request end

aryan-25:
Motivation:

In #44, we introduced a `fatalError` in `HTTPRequestConcludingAsyncReader` when the request stream ends before the request end part is received. This is incorrect behaviour, and the server should not fatal error in this scenario.

Modifications:

Replaced the `fatalError` with a `throw`.

Result:

The server no longer fatal errors when the request stream ends before the request end part is received.


Lukasa:
NB for the future: we should be adding regression tests when we fix things like this.

PR #55, opened 2026-02-09
Add support for swift-service-lifecycle

aryan-25:
Motivation:

The [`swift-service-lifecycle`](https://github.com/swift-server/swift-service-lifecycle) library is commonly used for managing service lifetimes. We should adopt `swift-service-lifecycle` for `NIOHTTPServer`.

Modifications:

- Updated the child channel initializers for HTTP/1.1 and Secure Upgrade in `NIOHTTPServer` to use [`swift-nio-extra`'s `ServerQuiescingHelper`](https://swiftpackageindex.com/apple/swift-nio-extras/main/documentation/nioextras/serverquiescinghelper). The `ServerQuiescingHelper` is added to the channel pipeline, and propagates a `ChannelShouldQuiesceEvent` upon graceful shutdown being triggered.

  - For HTTP/1.1, the [`HTTPServerPipelineHandler`](https://swiftpackageindex.com/apple/swift-nio/2.90.0/documentation/niohttp1/httpserverpipelinehandler) handler, which is contained in the pipeline, already handles `ChannelShouldQuiesceEvent` events.

  - For HTTP/2, a [`NIOHTTP2ServerConnectionManager`](https://swiftpackageindex.com/apple/swift-nio-http2/main/documentation/niohttp2/niohttp2serverconnectionmanagementhandler) is added to the pipeline. This connection manager reacts to a `ChannelShouldQuiesceEvent` and gracefully shuts down HTTP/2 connections.

- Updated `NIOHTTPServer`'s `serve(handler:)` to handle graceful shutdown (using `swift-service-lifecycle`'s [`withTaskCancellationOrGracefulShutdownHandler` method](https://swiftpackageindex.com/swift-server/swift-service-lifecycle/2.9.1/documentation/servicelifecycle/withtaskcancellationorgracefulshutdownhandler(isolation:operation:oncancelorgracefulshutdown:))).

Result:

Support for `swift-service-lifecycle` added.

### DIFF
--- a/Tests/NIOHTTPServerTests/NIOHTTPServer+ServiceLifecycleTests.swift
+++ b/Tests/NIOHTTPServerTests/NIOHTTPServer+ServiceLifecycleTests.swift
@@ -34,35 +34,35 @@ struct NIOHTTPServiceLifecycleTests {
     static let trailer: HTTPFields = [.trailer: "test_trailer"]
     static let reqEnd = HTTPRequestPart.end(trailer)
 
-    @Test("HTTP/1.1 in-flight request completes after graceful shutdown triggered")
+    let serverLogger = Logger(label: "Test Server")
+    let serviceGroupLogger = Logger(label: "Test ServiceGroup")
+
+    @Test("HTTP/1.1 active connection completes when graceful shutdown triggered", )
     @available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
-    func testHTTP1ConnectionInFlightRequestCompletesDuringGracefulShutdown() async throws {
+    func activeHTTP1ConnectionCanCompleteWhenGracefulShutdown() async throws {
         let server = NIOHTTPServer(
-            logger: Logger(label: "Test"),
+            logger: self.serverLogger,
             configuration: .init(bindTarget: .hostAndPort(host: "127.0.0.1", port: 0))
         )
 
-        // Create a promise that will be fulfilled when the server receives *part* of the request body. When this
-        // promise is fulfilled, we can initiate the graceful shutdown and then send the remaining body. If the server
-        // gracefully shuts down, we should be able to successfully complete the request.
+        // This promise will be fulfilled when the server receives the first part of the body. Once this happens, we can
+        // initiate the graceful shutdown and then send the remaining body. If graceful shutdown is respected, we should
+        // be able to successfully complete the request.
         let elg = MultiThreadedEventLoopGroup.singletonMultiThreadedEventLoopGroup
-        let partialRequestBodyReceivedPromise = elg.any().makePromise(of: Void.self)
+        let firstChunkReadPromise = elg.any().makePromise(of: Void.self)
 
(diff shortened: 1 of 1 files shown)


Sources/HTTPServer/ServiceLifecycle/HTTPService.swift
```diff
@@ -0,0 +1,61 @@
...
+    /// - Parameters:
+    ///   - server: The underlying HTTPServer instance.
+    ///   - serverHandler: The request handler that `server` will use.
+    ///   - onGracefulShutdown: A closure to execute upon graceful shutdown.
+    ///
+    /// - Note: The `onGracefulShutdown` closure will be called *after* initiating graceful shutdown on `server`.
+    public init(
+        server: Server,
+        serverHandler: Handler,
+        onGracefulShutdown gracefulShutdownHandler: @Sendable @escaping () -> Void = {}
```
FranzBusch: Why do we need to let the user customize this? They can setup their own handler by wrapping this service if they wanted to.

Package.swift
```diff
@@ -126,6 +139,8 @@ let package = Package(
             name: "NIOHTTPServerTests",
             dependencies: [
                 .product(name: "Logging", package: "swift-log"),
+                .product(name: "ServiceLifecycle", package: "swift-service-lifecycle"),
+                .product(name: "ServiceLifecycleTestKit", package: "swift-service-lifecycle"),
```
tib: Here you could also add the following: 

`condition: .when(traits: ["ServiceLifecycle"])`

aryan-25:
API breakage is expected: a new configuration argument was added.

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -149,15 +155,16 @@ public struct NIOHTTPServer: HTTPServer {
...
-            switch self.listeningAddressState.withLockedValue({ $0.close() }) {
-            case .failPromise(let promise, let error):
-                promise.fail(error)
-            case .doNothing:
-                ()
-            }
+        try await withTaskCancellationOrGracefulShutdownHandler {
+            try await self._serve(handler: handler)
+        } onCancelOrGracefulShutdown: {
+            self.close()
```
FranzBusch: This seems wrong to me. The behavior of a cancellation and graceful shutdown should be different. On cancellation we should force close all the channels on graceful shutdown we should quiesce

aryan-25: I've did some refactoring in order to get a reference to the underlying server channel in the `serve` method. It now looks like this:

```swift
let serverChannel = try await self.makeServerChannel()

return try await withTaskCancellationHandler {
    try await withGracefulShutdownHandler {
        try await self._serve(serverChannel: serverChannel, handler: handler)
    } onGracefulShutdown: {
        self.beginGracefulShutdown()
    }
} onCancel: {
    // Forcefully close down the server channel
    self.close(serverChannel: serverChannel)
}
```

FranzBusch (approved):
LGTM.

PR #58, opened 2026-02-18
Depend on `swift-http-api-proposal` and remove vendored-in code

gjcairo:
We had vendored-in `AsyncStreaming` types, `Middleware`, and all `HTTPServer`-related protocols from `swift-http-api-proposal` into this repo until that repo was public. Now that it is, we should move to depend on it.


gjcairo:
API break is expected.

PR #67, opened 2026-03-11
Refactor NIOHTTPServerConfiguration

aryan-25:
Motivation:

Currently, `NIOHTTPServerConfiguration`'s `transportSecurity` property implicitly determines the HTTP versions served by `NIOHTTPServer`:
- If `transportSecurity` is `.plaintext`, an HTTP/1.1 channel handler (over plaintext) is set up.
- If `transportSecurity` is `.tls` or `.mTLS` (or their certificate-reloading variants), an ALPN channel handler configured with both HTTP/1.1 and HTTP/2 handlers is set up.

As a consequence of this tight coupling, it is currently not possible to configure the server to only serve HTTP/1.1 over TLS.

Modifications:

- Added a new `supportedHTTPVersions` field (`Set<HTTPVersion>`) to `NIOHTTPServerConfiguration`. This, along with the `transportSecurity` value, allows for the following configurations to be represented:

  | `transportSecurity`  | `supportedHTTPVersions`     | Result                                                                                      |                                                      
  |----------------------|-----------------------------|---------------------------------------------------------|                                                                                                                             
  | `.plaintext`                  | `[.http1_1]`                             | HTTP/1.1 over plaintext                                                          |                                                                                                                             
  | `.tls` / `.mTLS`           | `[.http1_1]`                             | HTTP/1.1 over (m)TLS                                                            |
  | `.tls` / `.mTLS`           | `[.http2(...)]`                          | HTTP/2 over (m)TLS                                                              |
  | `.tls` / `.mTLS`           | `[.http1_1, .http2(...)]`           | HTTP/1.1 or HTTP/2 over (m)TLS, negotiated via ALPN.   |
  
- Simplified `TransportSecurity` to three options (`plaintext`, `tls`, `mTLS`):
  - TLS credentials (certificate chain and private key) are nested inside the `tls` and `mTLS` cases in a `TLSCredentials` type which has `inMemory`, `reloading`, and `pemFile` backings.
  - Similarly, mTLS trust configuration is represented through a new `MTLSTrustConfiguration` type which has `systemDefaults`, `inMemory`, `pemFile`, and `customCertificateVerificationCallback` backings.

- Updated `NIOHTTPServer` based on the changes to the configuration type.

- Updated the `swift-configuration` integration to reflect the new structure.

Result:

The HTTP versions supported by `NIOHTTPServer` can now be configured explicitly through the `supportedHTTPVersions` property on `NIOHTTPServerConfiguration`, rather than being implicitly determined by the transport security configuration.

### DIFF
--- a/Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
+++ b/Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
@@ -21,10 +21,14 @@ import SwiftASN1
 public import X509
 
 enum NIOHTTPServerConfigurationError: Error, CustomStringConvertible {
+    case customVerificationCallbackAndTrustRootsProvided
     case customVerificationCallbackProvidedWhenNotUsingMTLS
 
     var description: String {
         switch self {
+        case .customVerificationCallbackAndTrustRootsProvided:
+            "Invalid configuration. Both a custom certificate verification callback and a set of trust roots were provided. When a custom verification callback is provided, trust must be established directly within the callback."
+
         case .customVerificationCallbackProvidedWhenNotUsingMTLS:
             "Invalid configuration. A custom certificate verification callback was provided despite the server not being configured for mTLS."
         }
@@ -40,11 +44,15 @@ extension NIOHTTPServerConfiguration {
     /// ``NIOHTTPServerConfiguration`` is comprised of four types. Provide configuration for each type under the
     /// specified key:
     /// - ``BindTarget`` - Provide under key `"bindTarget"` (keys listed in ``BindTarget/init(config:)``).
+    ///
+    /// - ``SupportedHTTPVersions`` - Provide under key `"supportedHTTPVersions"` (keys listed in
+    ///   ``SupportedHTTPVersions/init(config:)``).
+    ///
     /// - ``TransportSecurity`` - Provide under key `"transportSecurity"` (keys listed in
     ///   ``TransportSecurity/init(config:customCertificateVerificationCallback:)``).
+    ///
     /// - ``BackPressureStrategy`` - Provide under key `"backpressureStrategy"` (keys listed in
     ///   ``BackPressureStrategy/init(config:)``).
-    /// - ``HTTP2`` - Provide under key `"http2"` (keys listed in ``HTTP2/init(config:)``).
     ///
     /// - Parameters:
     ///   - config: The configuration reader to read configuration values from.
@@ -64,12 +72,12 @@ extension NIOHTTPServerConfiguration {
 
         self.init(
             bindTarget: try .init(config: snapshot.scoped(to: "bindTarget")),
+            supportedHTTPVersions: try .init(config: snapshot),
             transportSecurity: try .init(
                 config: snapshot.scoped(to: "transportSecurity"),
                 customCertificateVerificationCallback: customCertificateVerificationCallback
             ),
-            backpressureStrategy: .init(config: snapshot.scoped(to: "backpressureStrategy")),
-            http2: .init(config: snapshot.scoped(to: "http2"))
+            backpressureStrategy: .init(config: snapshot.scoped(to: "backpressureStrategy"))
         )
     }
 }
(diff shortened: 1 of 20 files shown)


FranzBusch:
One configuration that this (and the current) configuration don't allow is to configure a single server with HTTP/1 over plaintext and then HTTP/2 over TLS for example. Not sure if we need to support it but might be worth thinking about if we need to allow a per-connection configuration somehow.

gjcairo:
@FranzBusch we considered this, but I'm not sure. Looking at Go's `http` package and Rust's `axum` (which I believe is widely used to run servers), as far as I can tell, you need to start separate servers for serving TLS and plaintext endpoints.

In Go, `http` offers [ListenAndServe](https://pkg.go.dev/net/http#ListenAndServe) and [ListenAndServeTLS](https://pkg.go.dev/net/http#ListenAndServeTLS):
```go
http.ListenAndServe(":8080", plaintextHandler)
http.ListenAndServeTLS(":443", "cert.pem", "key.pem", secureHandler)
```

For Axum it's similar: you either start plaintext servers via `axum::serve`, or you need to use `axum_server::bind_rustls(address, tlsConfig).serve(...)` ([see docs](https://docs.rs/axum-server/latest/axum_server/tls_rustls/index.html)).

Given this, I think it makes sense to have to create separate servers for plaintext vs TLS connections:

```swift
let plaintextServer = NIOHTTPServer(configuration: .init(transportSecurity: .plaintext))
let tlsServer = NIOHTTPServer(configuration: .init(transportSecurity: .tls(...)))

try await withThrowingDiscardingTaskGroup { group in
    group.addTask {
        try await plaintextServer.serve { request, requestContext, requestBodyAndTrailers, responseSender in
            // ...
        }
    }
    group.addTask {
        try await tlsServer.serve { request, requestContext, requestBodyAndTrailers, responseSender in
            // ...
        }
    }
}
```

Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
```diff
@@ -93,6 +101,49 @@ extension NIOHTTPServerConfiguration.BindTarget {
...
+
+        let versions = Set<HTTPVersionKind>(
+            try config.requiredStringArray(
+                forKey: "supportedHTTPVersions",
+                as: HTTPVersionKind.self
+            )
+        )
+
+        if versions.isEmpty {
+            fatalError("Invalid configuration: at least one supported HTTP version must be specified.")
```
gjcairo: I think we should be more consistent about throwing vs fatal-erroring. We throw errors in other scenarios, so we should probably do the same here.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
```diff
@@ -306,29 +236,41 @@ public struct NIOHTTPServerConfiguration: Sendable {
...
+        // If `transportSecurity`` is set to `.plaintext`, the server can only support HTTP/1.1. To support HTTP/2,
+        // `transportSecurity` must be set to `.tls` or `.mTLS`.
+        if case .plaintext = transportSecurity.backing, supportedHTTPVersions != [.http1_1] {
+            fatalError(
+                "Only HTTP/1.1 can be served over plaintext. transportSecurity must be set to (m)TLS for serving HTTP/2."
+            )
+        }
+
+        if supportedHTTPVersions.isEmpty {
+            fatalError("Invalid configuration: at least one supported HTTP version must be specified.")
```
gjcairo: I think we should throw errors here instead of fatal erroring. We should try to keep fatal errors for logic errors / theoretically unreachable states. However we could reach these states by user error (i.e. having invalid configs) so we should probably throw instead.

Tests/NIOHTTPServerTests/NIOHTTPServerSwiftConfigurationTests.swift
```diff
@@ -51,10 +51,9 @@ struct NIOHTTPServerSwiftConfigurationTests {
             let config = ConfigReader(provider: provider)
             let snapshot = config.snapshot()
 
-            let error = #expect(throws: Error.self) {
+            let configError = try #require(throws: Error.self) {
```
gjcairo, line 54: Have we got some more specific error type that gets thrown?

aryan-25: Unfortunately not, as [`ConfigError` is not public](https://github.com/apple/swift-configuration/blob/634baa6d9697024752dc2a8d0258b671095ee3fd/Sources/Configuration/ConfigReader.swift#L370).

gjcairo (approved):

FranzBusch (approved):

PR #68, opened 2026-03-20
Improve SwiftConfiguration integration

aryan-25:
Motivation:

PR #67 refactored `NIOHTTPServerConfiguration`. As part of this PR, we introduced support for TLS credentials and mTLS trust roots to be provided as PEM files too.

We should update the `swift-configuration` integration to match these changes. We should also use this opportunity to better document the possible keys and supported values in a DocC article, as the current documentation is spread inline across multiple types.

Modifications:

- Replaced the `"security"` key (5 values: `"plaintext"`, `"tls"`, `"reloadingTLS"`, `"mTLS"`, `"reloadingMTLS"`) with a `"transportSecurity.mode"` key (3 values: `"plaintext"`, `"tls"`, `"mTLS"`).
- Added a `"transportSecurity.credentialSource"` key (`"inline"`, `"file"`), and a `"transportSecurity.trustRootsSource"` key (`"inline"`, `"file"`, `"systemDefaults"`, `"customCertificateVerificationCallback"`).
- Renamed/rescoped several configuration keys for clarity (`"supportedHTTPVersions"` → `"http.versions"`, `"low"`/`"high"` → `"lowWatermark"`/`"highWatermark"`, scoped `"gracefulShutdown"`).
- Added a DocC article with a complete configuration key reference table, example JSON, and narrative documentation.

Result:

Improved `swift-configuration` integration.

### DIFF
--- a/Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
+++ b/Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
@@ -357,8 +357,8 @@ extension NIOHTTPServerConfiguration.HTTP2 {
     ///    connection.
     /// - `maxConcurrentStreams` (int, optional, default: nil): The maximum number of concurrent streams in an HTTP/2
     ///    connection.
-    /// - `gracefulShutdown.maximumGracefulShutdownDuration` (int, optional, default: nil): The maximum amount of time
-    ///    (in seconds) that the connection has to close gracefully.
+    /// - `gracefulShutdown.maximumDuration` (int, optional, default: nil): The maximum amount of time (in seconds) that
+    ///   the connection has to close gracefully.
     ///
     /// - Parameter config: The configuration reader.
     public init(config: ConfigSnapshotReader) {
@@ -385,13 +385,13 @@ extension NIOHTTPServerConfiguration.HTTP2.GracefulShutdownConfiguration {
     /// Initialize a HTTP/2 graceful shutdown configuration from a config reader.
     ///
     /// ## Configuration keys:
-    /// - `maximumGracefulShutdownDuration` (int, optional, default: nil): The maximum amount of time (in seconds) that
-    ///   the connection has to close gracefully.
+    /// - `maximumDuration` (int, optional, default: nil): The maximum amount of time (in seconds) that the connection
+    ///   has to close gracefully.
     ///
     /// - Parameter config: The configuration reader.
     public init(config: ConfigSnapshotReader) {
         self.init(
-            maximumGracefulShutdownDuration: config.int(forKey: "maximumGracefulShutdownDuration").map { .seconds($0) }
+            maximumGracefulShutdownDuration: config.int(forKey: "maximumDuration").map { .seconds($0) }
         )
     }
 }
--- a/Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
+++ b/Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
@@ -28,25 +28,26 @@ extension NIOHTTPServerConfiguration {
     ///
     /// ``NIOHTTPServerConfiguration`` is comprised of four types. Provide configuration for each type under the
     /// specified key:
-    /// - ``BindTarget`` - Provide under key `"bindTarget"` (keys listed in ``BindTarget/init(config:)``).
     ///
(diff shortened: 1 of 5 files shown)


Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
```diff
@@ -30,7 +30,7 @@ extension NIOHTTPServerConfiguration {
     /// specified key:
     /// - ``BindTarget`` - Provide under key `"bindTarget"` (keys listed in ``BindTarget/init(config:)``).
     ///
-    /// - ``SupportedHTTPVersions`` - Provide under key `"supportedHTTPVersions"` (keys listed in
+    /// - ``SupportedHTTPVersions`` - Provide under key `"http.versions"` (supported values listed in
```
czechboy0: Is this right? The input to the `init(config:)` of `SupportedHTTPVersions` is scoped to `http` only, the `versions` is then parsed internally and that type is `HTTPVersionKind` instead.

Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
```diff
@@ -374,8 +354,10 @@ extension NIOHTTPServerConfiguration.HTTP2 {
     /// - `maxFrameSize` (int, optional, default: 2^14): The maximum frame size to be used in an HTTP/2 connection.
     /// - `targetWindowSize` (int, optional, default: 2^16 - 1): The target window size to be used in an HTTP/2
     ///    connection.
-    /// - `maxConcurrentStreams` (int, optional, default: 100): The maximum number of concurrent streams in an HTTP/2
+    /// - `maxConcurrentStreams` (int, optional, default: nil): The maximum number of concurrent streams in an HTTP/2
     ///    connection.
+    /// - `gracefulShutdown.maximumGracefulShutdownDuration` (int, optional, default: nil): The maximum amount of time
```
czechboy0:
```suggestion
    /// - `gracefulShutdown.maximumDuration` (int, optional, default: nil): The maximum amount of time
```

Suggestion

Sources/NIOHTTPServer/Documentation.docc/SwiftConfigurationIntegration.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Configuration key reference
+
+``NIOHTTPServerConfiguration`` is comprised of four components. Provide the configuration for each component under its
+respective key prefix.
+
+> Important: HTTP/2 cannot be served over plaintext. If `"http2"` is included in `http.versions`, the transport
+> security must be set to `"tls"` or `"mTLS"`.
+
+| Prefix                        | Configuration Key                 | Type           | Required/Optional                                                                                                             | Default |
```
czechboy0, line 38: This table is a really good idea, I'll steal it for other projects too :)

czechboy0 (approved):
A few suggestions, otherwise lgtm

Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
aryan-25: Yes, sorry, `SupportedHTTPVersions` was a remnant from working on #67, but we eventually settled on using `Set<HTTPVersion>` to represent supported HTTP versions instead. I've updated the docs.

Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
aryan-25: Agreed.

I realised we also have a `maximumGracefulShutdownDuration` property under the `GracefulShutdownConfiguration` type in `NIOHTTPServerConfiguration`:

https://github.com/swift-server/swift-http-server/blob/b24399921e55760483140a11bd2293760108fb9b/Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift#L143

I'm going to address that in a separate PR where we will also make the other `NIOHTTP2ServerConnectionManagementHandler` options configurable; currently we only set the `maxGraceTime` option:
https://github.com/swift-server/swift-http-server/blob/b24399921e55760483140a11bd2293760108fb9b/Sources/NIOHTTPServer/NIOHTTPServer%2BSecureUpgrade.swift#L147-L151

gjcairo (approved):

PR #71, opened 2026-03-27
Fix request error handling

aryan-25:
Motivation:

We currently [propagate errors out of the `handleRequestChannel` method](https://github.com/swift-server/swift-http-server/blob/593b7caf20b26dbb7fd5f1676ca6b98a82f58ade/Sources/NIOHTTPServer/NIOHTTPServer.swift#L292).

For the plaintext HTTP/1.1 transport, the `handleRequestChannel` method is [called from within a `withThrowingDiscardingTaskGroup`](https://github.com/swift-server/swift-http-server/blob/593b7caf20b26dbb7fd5f1676ca6b98a82f58ade/Sources/NIOHTTPServer/NIOHTTPServer%2BHTTP1_1.swift#L33) without any error catching.

Since `withThrowingDiscardingTaskGroup` [cancels itself whenever *any* of the tasks spawned inside of it throws](https://developer.apple.com/documentation/swift/withthrowingdiscardingtaskgroup(returning:isolation:body:)#Error-Handling-and-Implicit-Cancellation), an error thrown in one connection brings the entire server down. We should fix this.

Modifications:

- Modified the behaviour of `handleRequestChannel` to close the channel when an error is caught instead of propagating the error upwards. 
- Added an associated test case.

Result:
 
An error thrown in a plaintext HTTP/1.1 connection no longer results in the parent task group being cancelled.

### DIFF
--- a/Sources/NIOHTTPServer/NIOHTTPServer.swift
+++ b/Sources/NIOHTTPServer/NIOHTTPServer.swift
@@ -289,7 +289,7 @@ public struct NIOHTTPServer: HTTPServer {
         } catch {
             self.logger.debug("Error thrown while handling connection: \(error)")
             // TODO: We need to send a response head here potentially
-            throw error
+            try? await channel.channel.close()
         }
     }
 
--- a/Tests/NIOHTTPServerTests/NIOHTTPServerTests.swift
+++ b/Tests/NIOHTTPServerTests/NIOHTTPServerTests.swift
@@ -552,7 +552,6 @@ struct NIOHTTPServerTests {
                         .connectToTestHTTP1Server(at: serverAddress)
 
                     try await secondClientChannel.executeThenClose { inbound, outbound in
-                        // Only send a request head; finish the stream immediately afterwards.
                         try await outbound.write(.head(.init(method: .post, scheme: "http", authority: "", path: "/")))
                         try await outbound.write(.body(Self.bodyData))
                         try await outbound.write(.end(nil))
--- a/Tests/NIOHTTPServerTests/NIOHTTPServerTests.swift
+++ b/Tests/NIOHTTPServerTests/NIOHTTPServerTests.swift
@@ -509,13 +509,88 @@ struct NIOHTTPServerTests {
             )
         }
     }
+
+    @available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
+    @Test("Server can still process other connections despite one failing")
+    func testServerCanContinueDespiteFailedConnection() async throws {
+        let server = try self.makePlaintextHTTP1Server()
+
+        let elg: EventLoopGroup = .singletonMultiThreadedEventLoopGroup
+        let firstRequestErrorCaught = elg.any().makePromise(of: Void.self)
+
+        try await Self.withServer(
+            server: server,
+            serverHandler: HTTPServerClosureRequestHandler { request, context, requestReader, responseSender in
+                do {
+                    try await Self.echoResponse(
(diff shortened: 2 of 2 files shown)


Tests/NIOHTTPServerTests/NIOHTTPServerTests.swift
```diff
@@ -509,14 +509,89 @@ struct NIOHTTPServerTests {
...
+                        outbound.finish()
+                    }
+
+                    try await firstRequestErrorCaught.futureResult.get()
+
+                    let secondClientChannel = try await ClientBootstrap(group: .singletonMultiThreadedEventLoopGroup)
+                        .connectToTestHTTP1Server(at: serverAddress)
+
+                    try await secondClientChannel.executeThenClose { inbound, outbound in
+                        // Only send a request head; finish the stream immediately afterwards.
```
gjcairo:
```suggestion
```

gjcairo (changes requested):

gjcairo (approved):
Thanks Aryan!

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -289,7 +289,7 @@ public struct NIOHTTPServer: HTTPServer {
         } catch {
             self.logger.debug("Error thrown while handling connection: \(error)")
```
fabianfett, line 290: not a change made in this pr, but we should never use string interpolation for log messages. the correct pattern is to put the error message into metadata.


```suggestion
            self.logger.debug("Error thrown while handling connection", metadata: ["error": "\(error)"])
```

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -289,7 +289,7 @@ public struct NIOHTTPServer: HTTPServer {
...
             // TODO: We need to send a response head here potentially
-            throw error
+            try? await channel.channel.close()
```
fabianfett, line 292: what has happened to the OG error? that is just swallowed here? I think we should at least add a comment, why we swallow it here.

fabianfett: also I think, we should log the error, if we can't close the connection.

Sources/NIOHTTPServer/NIOHTTPServer.swift
fabianfett, line 219: this should remove the throws keyword here, which will make usage of TaskGroups much more explicit.

Sources/NIOHTTPServer/NIOHTTPServer.swift
gjcairo: I think this was already changed in a different PR - in any case there's an open issue to audit these log messages.

Sources/NIOHTTPServer/NIOHTTPServer.swift
gjcairo: The change in this PR was a bit of a workaround to avoid the whole server coming down when an error is thrown. I will work on handling this properly - that's part of what the TODO right above is. I'll make sure we also log the error when closing.

Sources/NIOHTTPServer/NIOHTTPServer.swift
gjcairo: Yeah, this I missed, we should remove it - mind opening a small follow up @aryan-25 ?

aryan-25: Yes, I'm going to shortly create a PR to address @fabianfett's comments.

aryan-25:
Feedback from @fabianfett is addressed in #76.

PR #72, opened 2026-03-30
Add connection limiting and timeouts

gjcairo:
This PR adds connection-level backpressure and timeout support to the server. 

A new optional `maxConnections` field on `NIOHTTPServerConfiguration` lets users cap the number of concurrent connections: when the limit is reached, the server stops accepting new connections by gating NIO read events on the server channel until existing connections close. 

Alongside this, a new `ConnectionTimeouts` configuration provides idle, read-header, and-read body timeouts to evict slow or misbehaving connections, to mitigate attacks against the connection limit. Timeouts are enabled by default with reasonable values (60s/30s/60s respectively); the connection limit is opt-in. 

Both features work across HTTP/1.1 and HTTP/2, with timeout handlers split appropriately between connection and stream levels for HTTP/2.


Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
```diff
@@ -65,7 +65,9 @@ extension NIOHTTPServerConfiguration {
                 config: snapshot.scoped(to: "transportSecurity"),
                 customCertificateVerificationCallback: customCertificateVerificationCallback
             ),
-            backpressureStrategy: .init(config: snapshot.scoped(to: "backpressureStrategy"))
+            backpressureStrategy: .init(config: snapshot.scoped(to: "backpressureStrategy")),
+            maxConnections: snapshot.int(forKey: "maxConnections"),
+            connectionTimeouts: .init(config: snapshot.scoped(to: "connectionTimeouts"))
```
aryan-25: We should update the swift-configuration reference table in [Documentation.docc/SwiftConfigurationIntegration.md](https://github.com/swift-server/swift-http-server/blob/2c7367c30057ec6f8ccc9ed83ec1edc776649f1c/Sources/NIOHTTPServer/Documentation.docc/SwiftConfigurationIntegration.md) with these new options

Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
```diff
@@ -446,4 +448,29 @@ extension CertificateVerificationMode {
         }
     }
 }
+@available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
+extension NIOHTTPServerConfiguration.ConnectionTimeouts {
+    /// Initialize connection timeouts configuration from a config reader.
+    ///
+    /// ## Configuration keys:
+    /// - `idle` (int, optional, default: 60): Maximum time in seconds a connection can remain idle.
+    ///   Set to `null` to disable.
```
aryan-25: "Set to `null` to disable" isn't true; each of these options will get nil-coalesced to the default value if the key isn't specified in the config.

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -322,6 +322,64 @@ public struct NIOHTTPServer: HTTPServer {
             secureUpgradeChannel.channel.close(promise: nil)
         }
     }
+
+    /// Adds timeout handlers (idle, read header, read body) to a child channel pipeline.
+    ///
+    /// Only handlers for non-nil timeouts are installed. This is called for both
+    /// HTTP/1.1 per-connection channels and HTTP/2 per-stream channels.
+    func addTimeoutHandlers(to channel: any Channel) throws {
```
aryan-25: Should these methods be an extension on `Channel` instead?

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -322,6 +322,64 @@ public struct NIOHTTPServer: HTTPServer {
...
+        let timeouts = self.configuration.connectionTimeouts
+
+        if let idle = timeouts.idle {
```
aryan-25: Looks like these three `if` blocks are identical to the method bodies below. We can probably call into those methods here.

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -322,6 +322,64 @@ public struct NIOHTTPServer: HTTPServer {
...
+        }
+    }
+
+    /// Adds only idle timeout handlers to a channel. Used for HTTP/2 connection-level channels
+    /// where read header/body timeouts are handled per-stream.
+    func addIdleTimeoutHandlers(to channel: any Channel) throws {
+        if let idle = self.configuration.connectionTimeouts.idle {
+            let idleTimeAmount = TimeAmount(idle)
+            try channel.pipeline.syncOperations.addHandler(
+                IdleStateHandler(readTimeout: idleTimeAmount, writeTimeout: idleTimeAmount)
```
aryan-25: Should we allow both the read and write idle timeouts to be configurable?

Tests/NIOHTTPServerTests/ConnectionBackpressureEndToEndTests.swift
```diff
@@ -0,0 +1,214 @@
...
+                maxConnections: 2,
+                connectionTimeouts: .init(idle: nil, readHeader: nil, readBody: nil)
+            )
+        )
+
+        try await confirmation(expectedCount: 2) { responseReceived in
+            try await NIOHTTPServerTests.withServer(
+                server: server,
+                serverHandler: HTTPServerClosureRequestHandler { _, _, reader, responseSender in
+                    _ = try await reader.consumeAndConclude { bodyReader in
```
aryan-25: We can use [`NIOHTTPServerTests.echoResponse`](https://github.com/swift-server/swift-http-server/blob/2c7367c30057ec6f8ccc9ed83ec1edc776649f1c/Tests/NIOHTTPServerTests/NIOHTTPServerTests.swift#L725) here.

Tests/NIOHTTPServerTests/ConnectionBackpressureEndToEndTests.swift
```diff
@@ -0,0 +1,214 @@
...
+                                    group: .singletonMultiThreadedEventLoopGroup
+                                ).connectToTestHTTP1Server(at: serverAddress)
+
+                                try await client.executeThenClose { inbound, outbound in
+                                    try await outbound.write(
+                                        .head(.init(method: .get, scheme: "http", authority: "", path: "/"))
+                                    )
+                                    try await outbound.write(.end(nil))
+
+                                    var iter = inbound.makeAsyncIterator()
```
aryan-25: We can use [`NIOHTTPServerTests.validateResponse`](https://github.com/swift-server/swift-http-server/blob/2c7367c30057ec6f8ccc9ed83ec1edc776649f1c/Tests/NIOHTTPServerTests/NIOHTTPServerTests.swift#L613) here.

Sources/NIOHTTPServer/ConnectionLimitHandler.swift
```diff
@@ -0,0 +1,61 @@
...
+                    channel.read()
+                }
+            }
+        }
+
+        context.fireChannelRead(data)
+    }
+
+    func read(context: ChannelHandlerContext) {
+        if self.activeConnections <= self.maxConnections {
```
aryan-25, line 60: Is this channel handler guaranteed to only prevent the call to `read()` once?

Because `channel.read()` is only called once upon close (line 48).

Sources/NIOHTTPServer/ConnectionLimitHandler.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        let childChannel = self.unwrapInboundIn(data)
+        self.activeConnections += 1
+
+        let loopBoundSelf = NIOLoopBound(self, eventLoop: context.eventLoop)
+        let loopBoundContext = NIOLoopBound(context, eventLoop: context.eventLoop)
+        let eventLoop = context.eventLoop
+        childChannel.closeFuture.whenComplete { _ in
+            eventLoop.execute {
```
fabianfett, line 45: why do we need to hop here? we already should be on the correct EL.

Sources/NIOHTTPServer/ConnectionLimitHandler.swift
```diff
@@ -0,0 +1,62 @@
...
+                let `self` = loopBoundSelf.value
+                let context = loopBoundContext.value
+                self.activeConnections -= 1
+                if self.activeConnections <= self.maxConnections {
+                    context.read()
```
fabianfett, line 51: I don't think we should read here automatically. I think we should only read here, if we have seen a read call before that we didn't immediately forward. Other channels in the pipeline might want to stop backpressure for their own reasons. This auto read call assumes we are the only channel in the pipeline.

Sources/NIOHTTPServer/TimeoutHandlers.swift
```diff
@@ -0,0 +1,106 @@
...
+
+import NIOCore
+import NIOHTTPTypes
+
+/// A channel handler that closes the connection if the complete request headers
+/// are not received within the configured timeout.
+///
+/// The timeout starts when the channel becomes active and is cancelled when
+/// a `.head` part is received.
+final class ReadHeaderTimeoutHandler: ChannelInboundHandler, RemovableChannelHandler {
```
fabianfett: While it is totally fair to implement this in additional ChannelHanders, I strongly advise against it for performance reasons. We should really try to reduce pipeline overhead and pipeline modifications.

fabianfett (changes requested):

Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
gjcairo: I changed the docs but also removed the coalescing from the SwiftConfiguration initialiser - I think it's clearer that if you set things to nil then you don't want anything.

Sources/NIOHTTPServer/ConnectionLimitHandler.swift
gjcairo: Good point, will make this change

Sources/NIOHTTPServer/TimeoutHandlers.swift
gjcairo: I've consolidated these into two handlers instead. Is that what you had in mind?

Sources/NIOHTTPServer/ConnectionLimitHandler.swift
gjcairo: I might be misunderstanding your point, but the point of this handler is not to prevent the call to `read` just once - it's to make sure we're always below the set limit. That might mean calling read multiple times whenever we fall below the threshold, and withhold forwards of `read`s when we're above it.

Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
```diff
@@ -446,4 +448,23 @@ extension CertificateVerificationMode {
         }
     }
 }
+@available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
+extension NIOHTTPServerConfiguration.ConnectionTimeouts {
+    /// Initialize connection timeouts configuration from a config reader.
+    ///
+    /// ## Configuration keys:
+    /// - `idle` (int, optional, default: nil): Maximum time in seconds a connection can remain idle.
+    /// - `readHeader` (int, optional, default: nil): Maximum time in seconds to receive request headers.
```
fabianfett: after the connection/stream was open I assume. Can we please add that?

Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
```diff
@@ -446,4 +448,23 @@ extension CertificateVerificationMode {
...
+    /// - `readBody` (int, optional, default: nil): Maximum time in seconds to receive the request body.
```
fabianfett: the full body? or is this between body part? please make this more explicit.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
```diff
@@ -230,6 +230,51 @@ public struct NIOHTTPServerConfiguration: Sendable {
...
+        /// before being closed. `nil` means no idle timeout.
+        public var idle: Duration?
+
+        /// Maximum time allowed to receive the complete request headers
+        /// after a connection is established. `nil` means no timeout.
+        public var readHeader: Duration?
+
+        /// Maximum time allowed to receive the complete request body
+        /// after headers have been received. `nil` means no timeout.
+        public var readBody: Duration?
```
fabianfett, line 258: is this use-full? this means that this has to account for the longest possible request. I think something like the time between body parts is more appropriate?

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
```diff
@@ -230,6 +230,51 @@ public struct NIOHTTPServerConfiguration: Sendable {
...
 
+    /// Configuration for connection timeouts.
+    ///
+    /// Timeouts are enabled by default with reasonable values to protect against
+    /// slow or idle connections. Individual timeouts can be disabled by setting
+    /// them to `nil`.
+    public struct ConnectionTimeouts: Sendable {
+        /// Maximum time a connection can remain idle (no data read or written)
+        /// before being closed. `nil` means no idle timeout.
+        public var idle: Duration?
```
fabianfett, line 250: please call out, that this is only used after the first request. before the first request, it is `readHeader`.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
```diff
@@ -242,18 +287,31 @@ public struct NIOHTTPServerConfiguration: Sendable {
...
+    ///   - maxConnections: The maximum number of concurrent connections. `nil` means unlimited.
+    ///   - connectionTimeouts: The connection timeout configuration.
     public init(
         bindTarget: BindTarget,
         supportedHTTPVersions: Set<HTTPVersion>,
         transportSecurity: TransportSecurity,
-        backpressureStrategy: BackPressureStrategy = .defaults
+        backpressureStrategy: BackPressureStrategy = .defaults,
+        maxConnections: Int? = nil,
+        connectionTimeouts: ConnectionTimeouts = .defaults
```
fabianfett: are we sure we want to have this initializer? it feels like we would need to extend that, whenever a new property is added? I think having the init with just first three is more appropriate. All others can be changed via properties, if needed.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfigurationError.swift
```diff
@@ -24,6 +25,9 @@ enum NIOHTTPServerConfigurationError: Error, CustomStringConvertible {
 
         case .incompatibleTransportSecurity:
             "Invalid configuration: only HTTP/1.1 can be served over plaintext. `transportSecurity` must be set to (m)TLS for serving HTTP/2."
```
fabianfett, line 27: just as a drive by notice: I think we should also support, h/2 over plaintext, shouldn't we?

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -322,6 +322,41 @@ public struct NIOHTTPServer: HTTPServer {
             secureUpgradeChannel.channel.close(promise: nil)
         }
     }
+
+}
+
+@available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
+extension Channel {
```
fabianfett: i think this should be extension on the syncOperation and not the channel. This would make reading the callsite much easier.

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
```diff
@@ -120,6 +126,8 @@ extension NIOHTTPServer {
             channel.eventLoop.makeCompletedFuture {
                 try channel.pipeline.syncOperations.addHandler(HTTP1ToHTTPServerCodec(secure: true))
 
+                try channel.addTimeoutHandlers(self.configuration.connectionTimeouts)
```
fabianfett: the call should be:
```suggestion
                try channel.pipeline.syncOperations.addTimeoutHandlers(self.configuration.connectionTimeouts)
```

Sources/NIOHTTPServer/TimeoutHandlers.swift
```diff
@@ -0,0 +1,126 @@
...
+///
+/// This combines header and body read timeouts into a single handler with a
+/// state machine:
+/// - On channel active, a header timeout is scheduled (if configured).
+/// - When `.head` is received, the header timeout is cancelled and a body
+///   timeout is scheduled (if configured).
+/// - When `.end` is received, the body timeout is cancelled.
+///
+/// If either timeout fires, the connection is closed.
+final class RequestTimeoutHandler: ChannelInboundHandler, RemovableChannelHandler {
```
fabianfett, line 101: we 1000% should combine those two handlers into just one. We should try to reduce the total number of ChannelHandlers as much as possible.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfigurationError.swift
gjcairo: Yeah we're tracking that work.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
gjcairo: That's a fair question. Go's http package only has a single timeout from connection establishment till connection end (so, the time it takes to read everything, including headers and request body). 

I think having a timeout only between body parts could open the door to attacks where a client opens a bunch of connections and slowly trickles tiny bits of data just before the timeout fires to keep the connection alive. 

We could have both, but we already have the `idle` timeout which keeps track of all reads or writes, so it's somewhat covered. I suppose we could remove the idle timeout and instead add timeouts for the time between reads and a separate one for time between writes if we want more granularity (while also keeping this `readBody` timeout). What do you think?

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
gjcairo: This timeout runs from connection establishment and is reset every time something is read or written, it's independent of the readHeader timeout.

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
gjcairo: Well, it was correct before but yes, moving these methods to an extension of `SynchronousOperations` allows it to be spelled like this. I've changed it.

Sources/NIOHTTPServer/TimeoutHandlers.swift
gjcairo: I did it this way because for H2, we only want the idle handler in the connection channel, and the read handler in the stream channel, and merging them made it more confusing to read IMO as it would mean passing `nil` timeouts depending on which channel we're adding the handler to. I understand merging them would improve the performance of H1 requests though since we'd be adding a single handler instead of 2. I can make the change if we want to sacrifice some readability for performance in that case though.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
fabianfett: okay, so after connection establishment, there are two timers: `readHeader` and `idle`? Isn't this wasteful?

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
fabianfett: I'm in favor of the separate read and write timeouts. Because if the client doesn't accept any writes, we are also in a bad situation.

My question remains, how do you set a reasonable `readBody` timeout?

Also the `readBody` can easily be achieved using structured concurrency. We should consider if we really need this in the api.

fabianfett: This is btw. not a connectiontimeout but more a request timeout.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
```diff
@@ -230,6 +230,51 @@ public struct NIOHTTPServerConfiguration: Sendable {
...
+        public var readBody: Duration?
+
+        /// - Parameters:
+        ///   - idle: Maximum idle time before the connection is closed.
+        ///   - readHeader: Maximum time to receive request headers after a connection is established.
+        ///   - readBody: Maximum time to receive the complete request body after headers have been received.
+        public init(
+            idle: Duration? = Self.defaultIdle,
+            readHeader: Duration? = Self.defaultReadHeader,
+            readBody: Duration? = Self.defaultReadBody
```
fabianfett, line 267: if we add another timeout we have to add another initializer. Why not just add a `.init()` and users can set values using the vars?

Sources/NIOHTTPServer/TimeoutHandlers.swift
```diff
@@ -0,0 +1,126 @@
...
+
+    func handlerRemoved(context: ChannelHandlerContext) {
+        self.scheduledTimeout?.cancel()
+        self.scheduledTimeout = nil
+    }
+
+    private func scheduleTimeout(context: ChannelHandlerContext) {
+        self.scheduledTimeout?.cancel()
+        let boundContext = NIOLoopBound(context, eventLoop: context.eventLoop)
+        self.scheduledTimeout = context.eventLoop.scheduleTask(in: self.timeout) {
```
fabianfett: I think we should use `assumeIsolated()` here over NIOLoopBound. This allows us to only check the correct EL once. When calling `assumeIsolated()`

Sources/NIOHTTPServer/TimeoutHandlers.swift
fabianfett: > I can make the change if we want to sacrifice some readability for performance in that case though.

This would mean that we currently also translate H2 to H1 to add timeouts. I think we need specialized handlers here.

fabianfett (commented):
A new round of reviews.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
adam-fowler: Have you thought of doing a rate based check when reading the body. Expect a minimum number of bytes to have been received over a defined period of time.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
gjcairo: One is at the connection level and the other is at request level. For H1 I agree they overlap a bit, but for H2, these are pretty different things since once operates on the connection channel and the other operates at the stream channel.

Sources/NIOHTTPServer/Documentation.docc/SwiftConfigurationIntegration.md
```diff
@@ -57,6 +57,10 @@ respective key prefix.
 |                               | `certificateVerificationMode`     | `string`       | Required for `"mTLS"`, permitted values: `"optionalVerification"`, `"noHostnameVerification"`                                 | -       |
 | `backpressureStrategy`        | `lowWatermark`                    | `int`          | Optional                                                                                                                      | 2       |
 |                               | `highWatermark`                   | `int`          | Optional                                                                                                                      | 10      |
+| -                             | `maxConnections`                  | `int`          | Optional                                                                                                                      | nil     |
```
aryan-25: nit: there's a stray character here.
```suggestion
|                               | `maxConnections`                  | `int`          | Optional                                                                                                                      | nil     |
```

Sources/NIOHTTPServer/ConnectionLimitHandler.swift
gjcairo: The child channel's EL may be different from the parent's, that's why. I tried removing this anyways to confirm and saw a crash.

Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
```diff
@@ -483,4 +485,25 @@ extension CertificateVerificationMode {
...
+    /// after a connection is established.
+    /// - `readBody` (int, optional, default: nil): Maximum time in seconds to receive the complete request
+    /// body after headers have been received.
+    ///
+    /// - Parameter config: The configuration reader.
+    public init(config: ConfigSnapshotReader) {
+        self.init(
+            idle: config.int(forKey: "idle").map { .seconds($0) },
+            readHeader: config.int(forKey: "readHeader").map { .seconds($0) },
+            readBody: config.int(forKey: "readBody").map { .seconds($0) }
```
aryan-25, line 504: We should default these values to the same default values used in the normal `NIOHTTPServerConfiguration` initializer, given that its docs say this:
```
/// Timeouts are enabled by default with reasonable values to protect against
/// slow or idle connections. Individual timeouts can be disabled by setting
/// them to `nil`.
```

aryan-25 (approved):

Sources/NIOHTTPServer/TimeoutHandlers.swift
```diff
@@ -0,0 +1,143 @@
...
+    typealias OutboundOut = HTTPResponsePart
+
+    private let timeout: TimeAmount
+    private var scheduledTimeout: Scheduled<Void>?
+
+    init(timeout: TimeAmount) {
+        self.timeout = timeout
+    }
+
+    func channelActive(context: ChannelHandlerContext) {
```
FranzBusch, line 51: This might not be called if the handler is added and already active. You need to check in `channelHandlerAdded` if it is active and do the same you do here.

Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
gjcairo: Are you suggesting we default values when they're nil? How would you disable a timeout if we do this?

Sources/NIOHTTPServer/TimeoutHandlers.swift
gjcairo: Fixed

Sources/NIOHTTPServer/Configuration/NIOHTTPServer+SwiftConfiguration.swift
aryan-25: I see what you mean. Unless we add another boolean option like `readHeaderTimeoutEnabled`, there is no way to differentiate between "no timeout" vs "use default" when the key isn't present.

Adding a boolean option for each timeout will add unnecessary complexity to the config. I'm happy with the current shape.

PR #74, opened 2026-03-30
Add `Connection: close` header to responses temporarily

xbhatnag:
The server closes each client HTTP connection after writing out a single response. This confuses a client which expects to be able to pipeline multiple requests on the same connection.

To work around this, each response from the server will now include the `Connection: close` header informing the client that the connection will be closed after the response.

This causes issues with AsyncHTTPClient in particular, which does not expect this kind of non-compliance from servers.


xbhatnag:
Closing in favor of a proper fix that correctly handles pipelined requests on the server.

PR #76, opened 2026-04-01
Improve error handling

aryan-25:
Motivation:

The `serveInsecureHTTP1_1` and `serveSecureUpgrade` methods listen for incoming activity on the server channel and dispatch connections/requests to child tasks in order to process concurrently. However, these methods use a `withThrowingDiscardingTaskGroup`, and call many throwing functions from within the task group.

Since a `discarding` task group implicitly cancels *all* child tasks as soon as an error is encountered in *any* child task, we should carefully reason about and handle all errors that could arise within the task group.

Modifications:

- Addressed comments raised in #71. This included removing the redundant `throw` keyword from the `handleRequestChannel` declaration.

- Refactored `serveInsecureHTTP1_1` and `serveSecureUpgrade` (and associated methods) to handle errors more explicitly without propagating them up to the parent task group. As a result, it was then possible to replace `withThrowingDiscardingTaskGroup` with `withDiscardingTaskGroup`.

Result:

Better error handling.

### DIFF
--- a/Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
+++ b/Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
@@ -22,20 +22,38 @@ import NIOPosix
 
 @available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
 extension NIOHTTPServer {
+    /// Serves incoming plaintext HTTP/1.1 connections.
+    ///
+    /// Each connection is handled concurrently in its own child task. Individual connection errors are handled within
+    /// the child tasks and do not affect other connections.
+    ///
+    /// - Parameters:
+    ///   - serverChannel: The async channel that produces incoming HTTP/1.1 connections.
+    ///   - handler: The request handler.
+    ///
+    /// - Throws: If an error occurs while iterating the incoming connection stream.
     func serveInsecureHTTP1_1(
         serverChannel: NIOAsyncChannel<NIOAsyncChannel<HTTPRequestPart, HTTPResponsePart>, Never>,
         handler: some HTTPServerRequestHandler<RequestConcludingReader, ResponseConcludingWriter>
     ) async throws {
-        try await withThrowingDiscardingTaskGroup { group in
-            try await serverChannel.executeThenClose { inbound in
-                for try await http1Channel in inbound {
-                    group.addTask {
-                        try await self.handleRequestChannel(
-                            channel: http1Channel,
-                            handler: handler
-                        )
+        try await serverChannel.executeThenClose { inbound in
+            let inboundConnectionIterationError = await withDiscardingTaskGroup { group -> (any Error)? in
+                do {
+                    for try await http1Channel in inbound {
+                        group.addTask {
+                            await self.handleRequestChannel(channel: http1Channel, handler: handler)
+                        }
                     }
+
+                    return nil
+                } catch {
+                    return error
                 }
+            }
+
+            if let inboundConnectionIterationError {
+                // The error occurred while iterating the inbound connection stream
+                throw inboundConnectionIterationError
             }
         }
     }
--- a/Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
+++ b/Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
@@ -37,6 +37,10 @@ extension NIOHTTPServer {
         handler: some HTTPServerRequestHandler<RequestConcludingReader, ResponseConcludingWriter>
     ) async throws {
         try await serverChannel.executeThenClose { inbound in
+            // We don't use a `withThrowingDiscardingTaskGroup` here because an error thrown from the body or a child
+            // task would immediately propagate upwards, cancelling all child tasks and bringing down the entire server.
+            // We instead use a non-throwing discarding task group so that errors in the body (e.g. from iterating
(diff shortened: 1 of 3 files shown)


Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
```diff
@@ -22,21 +22,39 @@ import NIOPosix
...
-        try await withThrowingDiscardingTaskGroup { group in
-            try await serverChannel.executeThenClose { inbound in
-                for try await http1Channel in inbound {
-                    group.addTask {
-                        try await self.handleRequestChannel(
-                            channel: http1Channel,
-                            handler: handler
-                        )
+        try await serverChannel.executeThenClose { inbound in
+            let inboundConnectionIterationError = await withDiscardingTaskGroup { group -> (any Error)? in
```
gjcairo, line 44: Why can't we just use `withThrowingDiscardingTaskGroup` here? The only error we're throwing is the one coming from iterating `inbound`.

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
```diff
@@ -35,50 +35,116 @@ extension NIOHTTPServer {
...
-                                                    )
-                                                }
-                                            }
-                                        }
-                                    } catch {
-                                        self.logger.debug("HTTP2 connection closed: \(error)")
-                                    }
-                                }
+        try await serverChannel.executeThenClose { inbound in
+            let inboundConnectionIterationError = await withDiscardingTaskGroup { connectionGroup -> (any Error)? in
```
gjcairo, line 58: Similarly, I don't see why this is clearer than just using a throwing task group and throwing where appropriate?

Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
aryan-25: This is primarily to make it explicit that if you call a throwing function from within the body of the task group, the error must be handled directly rather than letting it propagate upwards and resulting in active tasks in the group being cancelled.

Currently iterating on `inbound` is the only source of a potential error, but if this was a `withThrowingDiscardingTaskGroup` then in the future we could inadvertently introduce a call to a throwing function without the compiler warning us.

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
aryan-25: Discussed [here](https://github.com/swift-server/swift-http-server/pull/76#discussion_r3022113708).

Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
gjcairo: Could we add some comment here (and in the other func) explaining this rationale? I am not convinced just doing things this way would prevent a future change from changing the behaviour unless it's more documented.

aryan-25: Agreed, addressed in commit cb84c52.

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -213,82 +213,101 @@ public struct NIOHTTPServer: HTTPServer {
...
-                        if !writerState.wrapped.withLock({ $0.finishedWriting }) {
-                            logger.error("Did not write response but error thrown.")
-                            // TODO: we need to do something, possibly just close the connection or
-                            // reset the stream with the appropriate error.
+                let httpRequest: HTTPRequest
+                switch nextPart {
+                case .head(let request):
+                    httpRequest = request
+                case .body:
+                    self.logger.debug("Unexpectedly received body on connection. Closing now")
```
fabianfett, line 247: Can we have something like a `connectionID` here? Can easily be done in a follow up...

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -213,82 +213,101 @@ public struct NIOHTTPServer: HTTPServer {
...
+                    outbound.finish()
+                    return
+                case .end:
+                    self.logger.debug("Unexpectedly received end on connection. Closing now")
```
fabianfett, line 251: Can we have something like a `connectionID` here? Can easily be done in a follow up...

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -213,82 +213,101 @@ public struct NIOHTTPServer: HTTPServer {
...
+
+                // Finish the outbound and wait on the close future to make sure all pending
+                // writes are actually written.
+                outbound.finish()
+                try await channel.channel.closeFuture.get()
+            }
         } catch {
-            self.logger.debug("Error thrown while handling connection: \(error)")
             // TODO: We need to send a response head here potentially
+            self.logger.error("Error thrown while handling connection", metadata: ["error": "\(error)"])
```
fabianfett, line 309: connectionID?

fabianfett (approved):
This looks much better. We might want to improve logging a bit, but we can do that in a follow up.

aryan-25:
Thanks. There's currently no mechanism to extract connection / stream information out of the underlying HTTP/1.1 or HTTP/2 connection channel. We'd need to add some methods to do that. I can address that in a follow-up PR.

PR #78, opened 2026-04-02
Fix multiple requests on same HTTP1 connection

gjcairo:
We're not currently handling keep alive HTTP1 connections properly: we close connections at the end of the first request, yet we don't send a connection close header. This PR fixes that by reusing the iterator until the stream is finished, so we can keep reading all requests that come through the same connection.

This also removes 6.2 support because there are some compiler issues with sendability that have not been properly addressed.

### DIFF
--- a/Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
+++ b/Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
@@ -139,7 +139,7 @@ final class HTTPKeepAliveHandler: ChannelDuplexHandler {
                 // The head we flushed earlier carried `Connection: close`; close
                 // the connection now that the response is complete.
                 context.flush()
-                context.close(mode: .all, promise: nil)
+                context.close(mode: .output, promise: nil)
             }
         }
     }
@@ -185,7 +185,7 @@ final class HTTPKeepAliveHandler: ChannelDuplexHandler {
             // The response was fully buffered (head + ... + end) and we have to
             // close. Close now (the flush above ensured the writes reached the
             // wire).
-            context.close(mode: .all, promise: nil)
+            context.close(mode: .output, promise: nil)
         }
     }
 }
--- a/Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
+++ b/Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
@@ -1,0 +1,171 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift HTTP Server open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift HTTP Server project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift HTTP Server project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import HTTPTypes
+import NIOCore
+import NIOHTTPTypes
+
+/// A NIO channel handler that ensures HTTP/1.1 keep-alive semantics are honored when
+/// the server starts writing a response before the request body has been fully read.
+///
+/// The handler buffers only the outbound response head whenever it is written before
+/// the request `.end` has been received. The buffered head is released in one of three
+/// ways:
+///
+/// - **Request `.end` arrives before any response part is written**: the head is
+///   flushed as-is and the response streams normally. The connection can be reused.
+/// - **A response body part is written before request `.end` arrives**: the buffered
+///   head is amended with `Connection: close`, then flushed; subsequent parts stream
+///   directly; once response `.end` is written, the connection is closed.
+/// - **Response `.end` is written while the head is still buffered (no body written)**:
+///   the head is amended with `Connection: close`, flushed, followed by `.end`, then
+///   the connection is closed.
+///
+/// Any time the head is flushed *because* the handler started producing the response
+/// before the request was fully read, the client receives `Connection: close` and the
+/// server itself closes the connection after writing response `.end`. This protects
+/// against clients that keep uploading request body bytes after the response has
+/// completed (which would otherwise force the server to drain unbounded data) and
+/// gives the client an explicit signal not to pipeline another request on the
+/// connection.
+///
+/// Informational (1xx) responses pass through unchanged and do not affect buffering
+/// state.
+@available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
+final class HTTPKeepAliveHandler: ChannelDuplexHandler {
+    typealias InboundIn = HTTPRequestPart
+    typealias InboundOut = HTTPRequestPart
+    typealias OutboundIn = HTTPResponsePart
+    typealias OutboundOut = HTTPResponsePart
+
+    private struct BufferedWrite {
+        var part: HTTPResponsePart
+        var promise: EventLoopPromise<Void>?
+    }
+
+    private enum FinalResponseState {
+        /// No final response has been written yet for the current request. Informational
+        /// (1xx) responses may have been passed through.
+        case notStarted
+        /// The final response head was written before request `.end` arrived. The head
+        /// is buffered until either request `.end` arrives (flush as-is, transition to
+        /// streaming, keep-alive), or a response body or `.end` is written (flush head
+        /// with `Connection: close`, transition to streaming, close after response
+        /// `.end`).
+        case bufferingHead(BufferedWrite)
+        /// The response is being streamed directly. If `closeAfterResponseEnd` is
+        /// true, the connection will be closed once response `.end` is written.
+        case streaming
+    }
+
+    /// `true` when the request `.end` has been received on the inbound side, or no
+    /// request is currently in flight. `false` between receiving a request `.head`
+    /// and its `.end`.
+    private var requestEndReceived: Bool = true
+
+    /// `true` if we've committed to closing the connection after this response's
+    /// `.end` is written. Set when the buffered head is flushed because a response
+    /// body or `.end` was written before request `.end` arrived. Cleared when a new
+    /// request begins.
+    private var closeAfterResponseEnd: Bool = false
+
+    private var finalResponseState: FinalResponseState = .notStarted
+
+    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        let part = self.unwrapInboundIn(data)
+        switch part {
+        case .head:
+            // Begin a new request. (Any previous request's response must have
+            // completed already since HTTPServerPipelineHandler enforces ordering.)
+            self.requestEndReceived = false
+            self.closeAfterResponseEnd = false
+            self.finalResponseState = .notStarted
+        case .body:
+            break
+        case .end:
+            self.requestEndReceived = true
+            // If we've been buffering the response head, flush it now: we can keep the
+            // connection alive.
+            if case .bufferingHead(let buffered) = self.finalResponseState {
+                self.finalResponseState = .streaming
+                context.write(self.wrapOutboundOut(buffered.part), promise: buffered.promise)
+                context.flush()
+            }
+        }
+        context.fireChannelRead(data)
+    }
+
+    func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+        let part = self.unwrapOutboundIn(data)
+        switch self.finalResponseState {
+        case .notStarted:
+            // Informational (1xx) responses pass through without affecting state: they
+            // don't conclude the response, so we remain in `.notStarted` until the
+            // final response head is written.
+            if case .head(let response) = part, response.status.kind == .informational {
+                context.write(data, promise: promise)
+                return
+            }
+            if self.requestEndReceived {
+                // Request fully read; stream the response directly.
+                self.finalResponseState = .streaming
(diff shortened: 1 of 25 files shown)


Sources/NIOHTTPServer/HTTPRequestConcludingAsyncReader.swift
```diff
@@ -119,20 +120,23 @@ public struct HTTPRequestConcludingAsyncReader: ConcludingAsyncReader, ~Copyable
...
+        /// Initializes a new request body reader.
         ///
-        /// - Parameter iterator: The NIO async channel inbound stream iterator to use for reading request parts.
-        fileprivate init(
-            iterator: consuming sending NIOAsyncChannelInboundStream<HTTPRequestPart>.AsyncIterator,
-            readerState: ReaderState
-        ) {
+        /// Takes the iterator from ReaderState so the state machine owns it for the entire read cycle.
+        fileprivate init(readerState: ReaderState) {
+            let iterator = readerState.takeIterator()
```
aryan-25: From what I understand, the iterator currently flows from `ReaderState` -> `RequestBodyAsyncReader` -> `RequestBodyStateMachine`.

`RequestBodyAsyncReader` does the following:
1. Takes the iterator out of `RequestState` (line 139)
2. Gives it to `RequestBodyStateMachine` (line 140)
3. Takes it out of `RequestBodyStateMachine` after getting a `readEnd` result from the state machine (line 168) and transfers it back to `ReaderState`.

If this is correct, should the `RequestBodyAsyncReader` initializer just pass `ReaderState` to `RequestBodyStateMachine`, and let `RequestBodyStateMachine` do the "taking" and "giving back" of the iterator directly with `ReaderState`?

I believe this would allow us to get rid of `takeIterator()` on `RequestBodyStateMachine` and move the "taking" and "giving back" parts closer to each other (it'd all happen in the state machine). I think this would improve local reasoning.

guoye-zhang:
I pointed https://github.com/apple/swift-http-api-proposal/pull/137 to the latest commit on this branch, but the AHC conformance test is still failing it appears

gjcairo:
@guoye-zhang hm, interesting. I'll take a look at that.

Sources/NIOHTTPServer/HTTPRequestConcludingAsyncReader.swift
gjcairo: Yeah that's a good suggestion, I've simplified it. Had to make some additional adjustments to get around sendability issues but it all seems to be working now

0xTim:
Just to but in here, this has fixed a load of test failures in Vapor that were failing with `Caught error: I/O on closed channel` so thank you!

gjcairo:
@guoye-zhang the failing conformance test now passes for me.

guoye-zhang:
I updated https://github.com/apple/swift-http-api-proposal/pull/137 and it's still running into these errors with AHC conformance tests: `✘ Test testOk failed with error: I/O on closed channel`

gjcairo:
`testOk` and `testBasicRedirect` are flaky, but they only fail like twice if I run them 100 times. 
Are you seeing them fail more consistently than this? Are we sure this is an issue on the server side?

guoye-zhang:
I'm actually reproducing it 100% on my Mac running the test in Xcode. GitHub CI reproduced it as well. Previous workaround with `Connection: close` never failed for me.

guoye-zhang (approved):

Sources/NIOHTTPServer/HTTPRequestConcludingAsyncReader.swift
```diff
@@ -88,21 +85,47 @@ public struct HTTPRequestConcludingAsyncReader: ConcludingAsyncReader, ~Copyable
...
                             fatalError("Unexpectedly received a request head.")
 
                         case .none:
-                            throw RequestBodyReadError.streamEndedBeforeReceivingRequestEnd
+                            // Stream ended without .end — don't put iterator back.
+                            throw RequestBodyReadError.requestEndedBeforeReceivingEnd
 
                         case .body(let element):
+                            nonisolated(unsafe) let iter = iterator
+                            self.readerState.putIterator(iter)
```
gjcairo: This is unfortunate because we need to go via the lock after each read - may slow things down considerably. Not sure what the alternative is other than not put any of this behind a lock, which I believe should be okay since we shouldn't be consuming the reader from more than one place at a time.

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -213,103 +213,107 @@ public struct NIOHTTPServer: HTTPServer {
...
-                        // TODO: we need to do something, possibly just close the connection or
-                        // reset the stream with the appropriate error.
+        // If the handler didn't fully consume the request body, drain the remaining
+        // parts so the iterator is positioned at the start of the next request.
+        // Errors during draining are not propagated — if the drain fails, we simply
+        // cannot reuse this connection.
+        if !readerState.wrapped.withLock({ $0.finishedReading }) {
+            do {
+                drainLoop: while true {
+                    switch try await recoveredIterator.next(isolation: #isolation) {
```
FranzBusch: We shouldn't do this. If the handler didn't fully consume the request body then we should close the connection in my opinion. Otherwise this could allow for attacks where an attacker crafts a request that leads to the handler responding early without having ready everything but keep the connection in an infinite draining state.

gjcairo: Hm yeah okay, you're right about the security concern. But wouldn't it potentially be a valid use case for a handler to respond early without consuming the whole request? The conformance tests actually do this for some requests: they don't read the request body (and thus they don't read `.end`) because they only look at the headers (or even nothing). Are we saying we should document that all handler implementations should make sure they always fully consume the request?

FranzBusch: I just took a quick look what Go does and they cap the drain to 256KB to stop bad actors. https://cs.opensource.google/go/go/+/master:src/net/http/server.go;l=1137?q=server.go&ss=go%2Fgo

@fabianfett WDYT?

gjcairo: I think this is reasonable. The only thing is that we can't really send back `Connection:close` by the time we run this drain loop (which is what Go does), because a response (and thus headers) may have already been sent back. We would just close the connection - which maybe we're okay with bur just pointing it out.

gjcairo: @FranzBusch I've implemented this change.

fabianfett: > may have already been sent back.

can we check, if they already have been sent back? if not add the `connection: close` header?

gjcairo: They will have been sent if the response sender was consumed, which happens whenever a Response is sent back to the client (by calling `responseSender.send(...)`. It would be odd to implement a handler that sends nothing back to the client and just returns; but yes, we could check whether `responseSender.send` has been called if we think that's useful.

gjcairo: We discussed this offline and decided that we would have a different approach: instead of the somewhat arbitrary 256kb payload drain, I've added an H1-only channel handler that adds the `Connection:close` header whenever we write a response `.end` or start sending back a response body when we haven't yet seen a request `.end`. After sending the request `.end` part, we'll close the connection.

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
```diff
@@ -0,0 +1,171 @@
...
+    private var closeAfterResponseEnd: Bool = false
+
+    private var finalResponseState: FinalResponseState = .notStarted
+
+    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        let part = self.unwrapInboundIn(data)
+        switch part {
+        case .head:
+            // Begin a new request. (Any previous request's response must have
+            // completed already since HTTPServerPipelineHandler enforces ordering.)
```
fabianfett, line 86: i think this is only true if pipelining is disabled.

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
```diff
@@ -0,0 +1,171 @@
...
+            // connection alive.
+            if case .bufferingHead(let buffered) = self.finalResponseState {
+                self.finalResponseState = .streaming
+                context.write(self.wrapOutboundOut(buffered.part), promise: buffered.promise)
+                context.flush()
+            }
+        }
+        context.fireChannelRead(data)
+    }
+
```
fabianfett, line 97: I think this also needs a timeout in the form of `channelReadComplete`. We should check the state in `channelReadComplete` and flush the head.

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
```diff
@@ -0,0 +1,171 @@
...
+/// server itself closes the connection after writing response `.end`. This protects
+/// against clients that keep uploading request body bytes after the response has
+/// completed (which would otherwise force the server to drain unbounded data) and
+/// gives the client an explicit signal not to pipeline another request on the
+/// connection.
+///
+/// Informational (1xx) responses pass through unchanged and do not affect buffering
+/// state.
+@available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
+final class HTTPKeepAliveHandler: ChannelDuplexHandler {
```
fabianfett, line 41: I really think we should unify the number of handlers here significantly. Doing this in a follow up is fine.

gjcairo: Sure, but I don't want to do it at this point - the timeouts are in a separate PR, and having them separate makes reviewing easier IMO. I will consolidate them in a follow up.

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
gjcairo: Hm, from the `HTTPServerPipelineHandler ` docs:

>This handler ensures that HTTP server pipelines only process one request at a time.

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
gjcairo: I'm trying to keep the timeouts in https://github.com/swift-server/swift-http-server/pull/72 - I haven't updated it since we had our chat though, I'll re-assign to you when I do.

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
fabianfett: https://github.com/apple/swift-nio/blob/6338e9fb5e136af8337b48eee65f8868d571a3d7/Sources/NIOHTTP1/HTTPPipelineSetup.swift#L232

yes, if we ensure that this is always enabled.

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
fabianfett: `channelReadComplete` isn't a timeout though :) It's another NIO callback.

gjcairo: I know, but then I don't think I understand your original comment :D

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
gjcairo: That option is enabled so we currently have this handler set up, which means this should be safe. We don't provide a way to change this configuration.

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
gjcairo: We discussed this offline: we agreed we would append `connection:close` to the response head if, by the time a read cycle ends (i.e. `channelReadComplete` is called) or `flush` is manually called, we haven't seen the request end.

fabianfett (approved):

FranzBusch (approved):
I think this is okay for now. I don't think that this design will hold for performance reasons. Furthermore, I don't think we could implement a similar thing with a non-NIO transport in the future.

gjcairo:
Regarding performance, you mean because of the additional channel handler and the intermediate buffering? 

The former will definitely be addressed: we should consolidate as many channel handlers as we can. For the buffering, maybe we could reuse the buffer? In any case I'd like to land this and work on perf in a follow up.

As for not being able to implement this in other implementations - each implementation can handle this differently, right? Why does it matter? Once we have a connection handle (which we could use to close a connection) I'm pretty sure this could even be implemented as a middleware if the concrete implementation didn't want to handle it for whatever reason.

FranzBusch:
> Regarding performance, you mean because of the additional channel handler and the intermediate buffering?
> 
> The former will definitely be addressed: we should consolidate as many channel handlers as we can. For the buffering, maybe we could reuse the buffer? In any case I'd like to land this and work on perf in a follow up.

Yes buffering + additional channel handler + the lock. Though the lock should be relatively free since it should be uncontended.

> As for not being able to implement this in other implementations - each implementation can handle this differently, right? Why does it matter? Once we have a connection handle (which we could use to close a connection) I'm pretty sure this could even be implemented as a middleware if the concrete implementation didn't want to handle it for whatever reason.

100%. Each implementation can handle this differently/

PR #79, opened 2026-05-13
Allow binding multiple addresses

gjcairo:
This PR adds support for binding a `NIOHTTPServer` to multiple addresses simultaneously. 

The configuration API now also accepts an array of `BindTarget`s. All resulting channels are served concurrently using the same request handler.

The server treats its listening addresses as a unit: if any channel stops serving for any reason, all remaining channels are are also cancelled. This behaviour matches the current behaviour with a single listening address.

The `listeningAddress` property has been replaced by `listeningAddresses`, an array of addresses.

### DIFF
--- a/Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
+++ b/Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
@@ -230,8 +230,8 @@ public struct NIOHTTPServerConfiguration: Sendable {
         }
     }
 
-    /// Network binding configuration
-    public var bindTarget: BindTarget
+    /// Network binding configuration specifying all addresses where the server should listen.
+    public var bindTargets: [BindTarget]
 
     /// TLS configuration for the server.
     public var transportSecurity: TransportSecurity
@@ -242,19 +242,23 @@ public struct NIOHTTPServerConfiguration: Sendable {
     /// Backpressure strategy to use in the server.
     public var backpressureStrategy: BackPressureStrategy
 
-    /// Create a new configuration.
+    /// Create a new configuration with multiple bind targets.
     /// - Parameters:
-    ///   - bindTarget: A ``BindTarget``.
+    ///   - bindTargets: An array of ``BindTarget`` values specifying where the server should listen.
     ///   - supportedHTTPVersions: The HTTP protocol versions the server should support.
     ///   - transportSecurity: The transport security mode (plaintext, TLS, or mTLS).
     ///   - backpressureStrategy: A ``BackPressureStrategy``.
     ///   Defaults to ``BackPressureStrategy/watermark(low:high:)`` with a low watermark of 2 and a high of 10.
     public init(
-        bindTarget: BindTarget,
+        bindTargets: [BindTarget],
         supportedHTTPVersions: Set<HTTPVersion>,
         transportSecurity: TransportSecurity,
         backpressureStrategy: BackPressureStrategy = .defaults
     ) throws {
+        if bindTargets.isEmpty {
+            throw NIOHTTPServerConfigurationError.noBindTargetsSpecified
+        }
+
         // If `transportSecurity`` is set to `.plaintext`, the server can only support HTTP/1.1.
         // To support HTTP/2, `transportSecurity` must be set to `.tls` or `.mTLS`.
         if case .plaintext = transportSecurity.backing {
@@ -267,10 +271,31 @@ public struct NIOHTTPServerConfiguration: Sendable {
             throw NIOHTTPServerConfigurationError.noSupportedHTTPVersionsSpecified
         }
 
-        self.bindTarget = bindTarget
+        self.bindTargets = bindTargets
         self.supportedHTTPVersions = supportedHTTPVersions
         self.transportSecurity = transportSecurity
         self.backpressureStrategy = backpressureStrategy
+    }
+
+    /// Create a new configuration with a single bind target.
+    /// - Parameters:
+    ///   - bindTarget: A ``BindTarget``.
+    ///   - supportedHTTPVersions: The HTTP protocol versions the server should support.
+    ///   - transportSecurity: The transport security mode (plaintext, TLS, or mTLS).
+    ///   - backpressureStrategy: A ``BackPressureStrategy``.
+    ///   Defaults to ``BackPressureStrategy/watermark(low:high:)`` with a low watermark of 2 and a high of 10.
+    public init(
+        bindTarget: BindTarget,
+        supportedHTTPVersions: Set<HTTPVersion>,
+        transportSecurity: TransportSecurity,
+        backpressureStrategy: BackPressureStrategy = .defaults
+    ) throws {
+        try self.init(
+            bindTargets: [bindTarget],
+            supportedHTTPVersions: supportedHTTPVersions,
+            transportSecurity: transportSecurity,
+            backpressureStrategy: backpressureStrategy
+        )
     }
 }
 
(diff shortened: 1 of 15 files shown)


Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
```diff
@@ -62,34 +62,41 @@ extension NIOHTTPServer {
...
-            try self.addressBound(serverChannel.channel.localAddress)
-
-            return serverChannel
+        var serverChannels = [NIOAsyncChannel<NIOAsyncChannel<HTTPRequestPart, HTTPResponsePart>, Never>]()
+        for bindTarget in bindTargets {
+            switch bindTarget.backing {
+            case .hostAndPort(let host, let port):
+                let serverChannel =
+                    try await bootstrap
+                    .bind(host: host, port: port) { channel in
```
aryan-25: nit: I think it'd be cleaner if this was in one line. Same applies in the secure upgrade method.

ehaydenr (approved):
Lgtm!

Sources/NIOHTTPServer/NIOHTTPServer+ListeningAddress.swift
```diff
@@ -37,25 +37,25 @@ enum ListeningAddressError: CustomStringConvertible, Error {
...
 
-    /// The address the server is listening from.
+    /// The addresses the server is listening on.
     ///
-    /// It is an `async` property because it will only return once the address has been successfully bound.
+    /// It is an `async` property because it will only return once the addresses have been successfully bound.
     ///
-    /// - Throws: An error will be thrown if the address could not be bound or is not bound any longer because the
+    /// - Throws: An error will be thrown if the addresses could not be bound or are not bound any longer because the
     ///   server isn't listening anymore.
```
aryan-25: Worth noting here that *all* addresses have to be successfully bound in order for this property to return.

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -116,100 +116,106 @@ public struct NIOHTTPServer: HTTPServer {
...
+    ///     )
     /// )
+    ///
+    /// try await server.serve(handler: MyHandler())
     /// ```
     public func serve(
         handler: some HTTPServerRequestHandler<RequestConcludingReader, ResponseConcludingWriter>
     ) async throws {
-        let serverChannel = try await self.makeServerChannel()
+        let serverChannels = try await self.makeServerChannels()
```
aryan-25: We bind each channel to the listening addresses in this method. Does this method throw if the bind failed for a particular channel?

I'm wondering if we can reach a state where some channels are bound to an address and some are not; we should make sure that the channels which have already started are then closed in this case.

Tests/NIOHTTPServerTests/NIOHTTPServerTests.swift
```diff
@@ -568,6 +568,170 @@ struct NIOHTTPServerTests {
...
+            }
+        )
+    }
+
+    /// Verifies the all-or-nothing listening semantics: when the server stops (e.g., due to cancellation),
+    /// all bound addresses become unavailable simultaneously and ``listeningAddresses`` throws
+    /// ``ListeningAddressError/serverClosed``. No subset of addresses continues serving after the server
+    /// has stopped.
+    @available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
+    @Test("All addresses stop together and listeningAddresses throws after server stops", .timeLimit(.minutes(1)))
```
aryan-25: Was `.timeLimit(.minutes(1))` intentional?

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
```diff
@@ -242,19 +242,23 @@ public struct NIOHTTPServerConfiguration: Sendable {
     /// Backpressure strategy to use in the server.
     public var backpressureStrategy: BackPressureStrategy
 
-    /// Create a new configuration.
+    /// Create a new configuration with multiple bind targets.
```
aryan-25, line 245: We'll need to update the `swift-configuration` initializer (and the markdown reference table) too now that multiple bind targets are supported.

Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
gjcairo: Sadly the formatter complains when it's all in one line

aryan-25: Is it the line length that's causing the issue?
 
I just tried this and the formatter seems to be fine with it:
```suggestion
                    try await bootstrap.bind(host: host, port: port) { channel in
```

This is way below our [configured line length](https://github.com/swift-server/swift-http-server/blob/446a2783ccadacec8c8ef2051926c8dbe231ef33/.swift-format#L17), which is 120.

Tests/NIOHTTPServerTests/NIOHTTPServerTests.swift
gjcairo: No it was a leftover, good catch

Sources/NIOHTTPServer/NIOHTTPServer.swift
gjcairo: Yeah good catch, we should close the channels. I'd assumed dropping the channels would trigger closes but I forgot we deprecated the closeOnDeinit by default on `NIOAsyncChannel`. I've fixed this.

Tests/NIOHTTPServerTests/NIOHTTPServerTests.swift
```diff
@@ -568,6 +568,233 @@ struct NIOHTTPServerTests {
...
+                    .hostAndPort(host: "127.0.0.1", port: 0),
+                    .hostAndPort(host: "127.0.0.1", port: 0),
+                ],
+                supportedHTTPVersions: [.http1_1],
+                transportSecurity: .plaintext
+            )
+        )
+
+        // Start the server, verify both addresses are listening, then cancel.
+        try await withThrowingTaskGroup { group in
```
aryan-25: We can use `withServer` here

Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
```diff
@@ -62,34 +62,48 @@ extension NIOHTTPServer {
...
+                                )
+                            )
+                        }
+                    serverChannels.append(serverChannel)
+                }
+            }
+        } catch {
+            // A later bind failed: close any channels we already bound to avoid leaking sockets.
+            for serverChannel in serverChannels {
+                serverChannel.channel.close(promise: nil)
```
aryan-25: Given that this method is `async`, I think we should `await` the close here:

```suggestion
                try? await serverChannel.channel.close()
```

Same applies for the secure upgrade method.

Tests/NIOHTTPServerTests/NIOHTTPServerTests.swift
```diff
@@ -568,6 +568,233 @@ struct NIOHTTPServerTests {
...
+    ///
+    @available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
+    @Test("Previously bound channels are closed when a later bind fails")
+    func testPreviouslyBoundChannelsAreClosedOnPartialBindFailure() async throws {
+        // We use a specific port for the first target rather than `port: 0`, because the
+        // "bind to ephemeral port 0, close, then verify the port is free later" pattern is racy in
+        // a parallel test environment: another test that binds to port 0 may be assigned the same
+        // port between the close and our verification. The chosen port is also below the typical
+        // ephemeral range so it cannot be allocated to other parallel tests using `port: 0`.
+        let firstPort = 30_210
```
aryan-25: We should be able to get rid of this and the `Task.sleep(for: .milliseconds(100))` below in line 787 if we await the `close`: see https://github.com/swift-server/swift-http-server/pull/79#discussion_r3258327740.

aryan-25 (approved):
Some small suggestions, LGTM otherwise :)

FranzBusch (approved):

PR #80, opened 2026-05-19
Move to new `AsyncStreaming` APIs

gjcairo:
This PR updates the `HTTPRequestConcludingAsyncReader` and `HTTPResponseConcludingAsyncWriter` to use the new `_AsyncStreaming` module in `swift-async-algorithms`, which has already been adopted in the abstract APIs in `swift-http-api-proposal`.

Some tests had to be modified and some deleted because the behaviour/semantics of the reader have changed.

We're now also 6.4+ only, as it's required for this new module.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:6.2
+// swift-tools-version:6.4
 //===----------------------------------------------------------------------===//
 //
 // This source file is part of the Swift HTTP Server open source project
--- a/Package.swift
+++ b/Package.swift
@@ -49,7 +49,7 @@ let package = Package(
     dependencies: [
         .package(
             url: "https://github.com/apple/swift-http-api-proposal.git",
-            branch: "c195a5dd2c05780908d2d29bb064f832e108eb5d"
+            revision: "c195a5dd2c05780908d2d29bb064f832e108eb5d"
         ),
         .package(
             url: "https://github.com/apple/swift-async-algorithms.git",
--- a/Package.swift
+++ b/Package.swift
@@ -49,7 +49,12 @@ let package = Package(
     dependencies: [
         .package(
             url: "https://github.com/apple/swift-http-api-proposal.git",
-            branch: "2cb0ef6722e2086a04b5a14c0c40971b8038099a"
+            branch: "c195a5dd2c05780908d2d29bb064f832e108eb5d"
+        ),
+        .package(
+            url: "https://github.com/apple/swift-async-algorithms.git",
+            revision: "d0b4a06d0f173a2f3be27d3ea21b3c3aa18db440",
+            traits: ["UnstableAsyncStreaming"]
         ),
         .package(url: "https://github.com/apple/swift-http-types.git", from: "1.0.0"),
         .package(url: "https://github.com/apple/swift-distributed-tracing.git", from: "1.0.0"),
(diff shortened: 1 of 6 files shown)


Package.swift
```diff
@@ -42,7 +49,12 @@ let package = Package(
     dependencies: [
         .package(
             url: "https://github.com/apple/swift-http-api-proposal.git",
-            branch: "2cb0ef6722e2086a04b5a14c0c40971b8038099a"
+            branch: "c195a5dd2c05780908d2d29bb064f832e108eb5d"
```
aryan-25:
```suggestion
            revision: "c195a5dd2c05780908d2d29bb064f832e108eb5d"
```

Sources/NIOHTTPServer/HTTPRequestConcludingAsyncReader.swift
```diff
@@ -38,93 +40,14 @@ public struct HTTPRequestConcludingAsyncReader: ConcludingAsyncReader, ~Copyable
         /// The type of errors that can occur during reading operations.
         public typealias ReadFailure = any Error
 
+        /// The buffer type used to hand elements to the caller.
+        public typealias Buffer = UniqueArray<UInt8>
```
FranzBusch, line 44: What if we add a conformance of `ByteBuffer` to `RangeReplaceableContainer` in this package as a `@retroactive` conformance for now. This would allow us to avoid any of the copies and allocs

Sources/NIOHTTPServer/HTTPRequestConcludingAsyncReader.swift
```diff
@@ -133,42 +56,42 @@ public struct HTTPRequestConcludingAsyncReader: ConcludingAsyncReader, ~Copyable
...
-                case .readEnd(let trailers):
-                    self.state.wrapped.withLock { state in
-                        state.trailers = trailers
-                        state.finishedReading = true
-                    }
-                    return try await body(.init())
-
-                case .streamFinished:
-                    return try await body(.init())
+            var buffer = UniqueArray<UInt8>()
```
FranzBusch: We should allocate that buffer once and re-use it. If we do the `ByteBuffer` conformance then that would be irrelevant.

Sources/NIOHTTPServer/HTTPResponseConcludingAsyncWriter.swift
```diff
@@ -39,7 +40,10 @@ public struct HTTPResponseConcludingAsyncWriter: ConcludingAsyncWriter, ~Copyabl
         public typealias WriteElement = UInt8
 
         /// The type of errors that can occur during writing operations.
-        public typealias WriteFailure = Error
+        public typealias WriteFailure = any Error
+
+        /// The buffer type used to receive elements from the caller.
+        public typealias Buffer = UniqueArray<UInt8>
```
FranzBusch, line 46: Same comment here

Sources/NIOHTTPServer/HTTPRequestConcludingAsyncReader.swift
gjcairo: Would this conformance be something that we want in NIO?

FranzBusch: At some point yes but I don't think we should add it there right now. I would propose adding it here for the time being as a `@retroactive` conformance

Sources/NIOHTTPServer/HTTPRequestConcludingAsyncReader.swift
```diff
@@ -133,42 +61,43 @@ public struct HTTPRequestConcludingAsyncReader: ConcludingAsyncReader, ~Copyable
...
+            self.buffer.removeAll(keepingCapacity: true)
+            switch requestPart {
+            case .head:
+                fatalError()
+            case .body(let element):
+                self.buffer.reserveCapacity(element.readableBytes)
+                unsafe element.withUnsafeReadableBytes { rawBufferPtr in
+                    let usbptr = unsafe rawBufferPtr.assumingMemoryBound(to: UInt8.self)
+                    unsafe self.buffer.append(copying: usbptr)
                 }
```
FranzBusch: You should be able to call `buffer.append(copying: element.span)`

FranzBusch (approved):

PR #89, opened 2026-06-13
Adopt final element and request context from swift-http-api-proposal 0.1.0

guoye-zhang:


Package.swift
```diff
@@ -49,43 +50,33 @@ let package = Package(
     dependencies: [
         .package(
             url: "https://github.com/apple/swift-http-api-proposal.git",
-            revision: "d58fd6fa157e08bff44aa360ff83ebd424783392"
+            revision: "c12fdd4c48953a691b1ce52357101e844e5f0887"
```
ehaydenr, line 53: Are we safe to point at 0.1.0? Or should we continue to update the revision pin like this?

guoye-zhang: I tried but Swift doesn't allow it since swift-http-api-proposal still uses commit hashes

gjcairo (approved):

PR #92, opened 2026-06-30
Add connection lifecycle handling

gjcairo:
## Summary

Add a connection-lifecycle surface to `NIOHTTPServer` so users can do per-connection setup (such as loggers, metrics, atomic counters, etc) and share state across requests on the same connection. The existing `serve(handler:)` entry point is preserved unchanged at the call site.

## What's new

- **`NIOHTTPServer.serve(connectionHandler:)`**: the connection-aware counterpart to `serve(handler:)`. 
- **`NIOHTTPServerConnectionHandler`** protocol with `handleConnection(connection:context:)`.
- **`NIOHTTPServer.Connection`**: struct that represents an active HTTP server connection. Consumed by exactly one call to `handleRequests`; if the handler returns without calling `handleRequests`, the connection is dropped cleanly on scope exit.
- **`NIOHTTPServer.ConnectionContext`**, which now exposes `httpVersion`, `remoteAddress`, `localAddress`, the existing `peerCertificateChain`, and a new `signalConnectionClose()`.
- **New `HTTPServerCapability` capabilities:** `ConnectionInfo`, `PeerCertificate`. `NIOHTTPServer.RequestContext` is now a concrete struct conforming them, so request handlers reach connection-scoped data via `requestContext.remoteAddress`, etc. 
- The `@TaskLocal NIOHTTPServer.connectionContext` API is removed, and all access is now through `requestContext`.


Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
```diff
@@ -0,0 +1,61 @@
...
+
+@available(anyAppleOS 26.0, *)
+extension HTTPServerCapability {
+    /// A request-context capability exposing connection-scoped peer and local addresses.
+    ///
+    /// Servers whose request context conforms to this capability surface the
+    /// peer's address and the local address the connection is bound to. Both
+    /// are reported best-effort: implementations may return `nil` when the
+    /// underlying transport cannot report an address.
+    public protocol ConnectionInfo: RequestContext {
```
FranzBusch, line 26: I think we really wanna add that to the `HTTPAPIs`. Can you file an issue there?

Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
```diff
@@ -0,0 +1,61 @@
...
+        var localAddress: NIOHTTPServer.SocketAddress? { get }
+    }
+
+    /// A request-context capability exposing the validated peer certificate chain.
+    ///
+    /// Servers whose request context conforms to this capability surface the
+    /// peer's mTLS-validated certificate chain (when applicable). Implementations
+    /// return `nil` if mTLS isn't configured, or if no validated chain was
+    /// derived.
+    public protocol PeerCertificate: RequestContext {
```
FranzBusch, line 40: Same for this

Sources/ConnectionHandlerExample/ConnectionHandlerExample.swift
```diff
@@ -0,0 +1,89 @@
...
+        try await server.serve { connection, context in
+            let connectionLogger: Logger = {
+                var logger = rootLogger
+                let peer = context.remoteAddress.map { "\($0)" } ?? "unknown"
+                logger[metadataKey: "peer"] = .string(peer)
+                logger[metadataKey: "http"] = .string("\(context.httpVersion)")
+                return logger
+            }()
+            connectionLogger.info("connection accepted")
+            defer { connectionLogger.info("connection closed") }
```
FranzBusch: Let's use the new task local Logging APIs for this example. Actually the HTTP Server itself should set the task local logger and configure it correctly with metadata. This shouldn't be up to the user.

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
FranzBusch, line 1: Can we keep the docs updates like this separate?

Sources/NIOHTTPServer/NIOHTTPServer+ConnectionContext.swift
```diff
@@ -12,25 +12,78 @@
...
+import NIOConcurrencyHelpers
 import NIOCore
 import NIOSSL
 public import X509
 
 @available(anyAppleOS 26.0, *)
 extension NIOHTTPServer {
-    /// Connection-specific information available during request handling.
+    /// The application-level HTTP version negotiated for a connection.
+    public enum HTTPVersion: Sendable, Hashable {
```
FranzBusch: This should also become part of HTTPAPIs

Sources/ConnectionHandlerExample/ConnectionHandlerExample.swift
ehaydenr: what metadata should be included? In some applications it may not be desirable to log certain things like peer IP.

Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
ehaydenr: What would be used as the SocketAddress type in HTTPAPIs?

ehaydenr (commented):
This is looking good at a high level. Can we split it up into two PRs, though? I think CloseableConnection might make sense to have as a subsequent PR.

Sources/ConnectionHandlerExample/ConnectionHandlerExample.swift
FranzBusch: The HTTP server should make a decision about the general useful metadata it wants to include. I would say things like HTTP version and peer IP are sensible. The rest the user can extract from the context and add.

Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
FranzBusch: We need to add a new SocketAddress currency type in the HTTPAPIs that we need to land in the planned network currency types package

Sources/ConnectionHandlerExample/ConnectionHandlerExample.swift
ehaydenr: IMO, peer IP is sensitive and shouldn't be logged by default

Sources/NIOHTTPServer/NIOHTTPServer+ConnectionContext.swift
gjcairo: https://github.com/apple/swift-http-api-proposal/issues/180

Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
gjcairo: https://github.com/apple/swift-http-api-proposal/issues/179

Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
gjcairo: https://github.com/apple/swift-http-api-proposal/issues/178

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
gjcairo: Even when it's related to the change being introduced in the PR?

gjcairo:
@ehaydenr split it out: https://github.com/swift-server/swift-http-server/pull/94

Sources/NIOHTTPServer/NIOHTTPServer+Connection.swift
```diff
@@ -0,0 +1,124 @@
...
+            self.context = context
+            self.httpProtocol = httpProtocol
+        }
+
+        /// Run the request loop on this connection until it terminates.
+        ///
+        /// Each request received on this connection is dispatched to `handler`. The
+        /// loop returns when the peer closes the connection, the server shuts down,
+        /// or an error occurs.
+        public consuming func handleRequests<Handler: HTTPServerRequestHandler>(
```
ehaydenr, line 67: Can we return why the connection closed? For example, was there an error? Who initiated the closure? At the end of a connection, an application may want to ask questions about what happened or retrieve summary data. I think `consuming` is probably right here, but it prevents the app from being able to do that - which is why I think a return type probably makes sense. In a sketch of this I did, I used a ConnectionSummary struct which contained all this information.

ehaydenr: Discussed offline. I'm fine with this being a follow up

Sources/NIOHTTPServer/NIOHTTPServer+Connection.swift
```diff
@@ -0,0 +1,124 @@
...
+            handler: Handler
+        ) async throws
```
aryan-25: There is nothing in this method that `throws`.

Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
```diff
@@ -70,6 +66,45 @@ extension NIOHTTPServer {
...
+        do {
+            try await requestChannel.executeThenClose { inbound, outbound in
+                let context = NIOHTTPServer.makeHTTP1ConnectionContext(
+                    requestChannel: requestChannel,
+                    peerCertificateChainFuture: nil
+                )
+                let connection = Connection(
+                    server: self,
+                    context: context,
+                    httpProtocol: .http1_1(inbound: inbound, outbound: outbound)
```
aryan-25, line 89: I think it would be better if we pass the `requestChannel` _itself_ as the associated value instead of calling `executeThenClose` here.

`handleHTTP1RequestLoop` itself has a `do-catch` block. This will let us remove the unnecessary outer do-catch block in this method.

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
```diff
@@ -105,66 +78,114 @@ extension NIOHTTPServer {
...
+                    } catch {
+                        self.logger.debug(
+                            "Error thrown by connection handler",
+                            metadata: ["error": "\(error)"]
+                        )
+                    }
+                }
+            } catch {
+                self.logger.debug(
+                    "Error tearing down HTTP/1.1 channel",
```
aryan-25: `executeThenClose` doesn't throw if there was an error _closing_ the channel. It throws the error that was caught in the body, so the "tearing down" part of the log statement isn't accurate here. 

See https://github.com/apple/swift-nio/blob/b1f2dd18fcfd3f4b0a6d7ddc55c683c494b085a3/Sources/NIOCore/AsyncChannel/AsyncChannel.swift#L333-L372.

Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
aryan-25: Edit: Ignore this comment. I ran the tests and realised we need this to avoid the NIOAsyncWriter deinit precondition if an error is thrown before calling `handleRequests`.

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
```diff
@@ -105,66 +78,114 @@ extension NIOHTTPServer {
...
+                context: context,
+                httpProtocol: .http2(connectionChannel: connectionChannel, multiplexer: multiplexer)
             )
+            do {
+                try await connectionHandler.handleConnection(connection: connection, context: context)
+            } catch {
+                self.logger.debug(
+                    "Error thrown by connection handler",
+                    metadata: ["error": "\(error)"]
+                )
```
aryan-25, line 145: We should add a `try? await connectionChannel.close()` here.

Otherwise, the connection does not get cleaned up if an error is thrown before `handleRequests` is called.

aryan-25: Edit: `try? await connectionChannel.close()` should be after the do-catch block so that we also clean up the connection if `handleConnection` did not call `handleRequests`.

Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
```diff
@@ -0,0 +1,44 @@
...
+        /// The peer's validated certificate chain, when available.
+        var peerCertificateChain: X509.ValidatedCertificateChain? { get async throws }
```
FranzBusch, line 42: Why is this `async` and `throws`? That seems very odd

gjcairo: This was already its signature - I'm just introducing this capability. Look at `NIOHTTPServer+ConnectionContext.swift`:`ConnectionContext/peerCertificateChain`.
Based on how we're getting it, it has to be async throws.

Sources/NIOHTTPServer/NIOHTTPServerDefaultConnectionHandler.swift
```diff
@@ -41,6 +41,6 @@ where
         connection: consuming sending NIOHTTPServer.Connection,
         context: NIOHTTPServer.ConnectionContext
     ) async throws {
```
aryan-25, line 43: We can remove `throws` from here now (and in all the `NIOHTTPServerConnectionHandler` conformances in the tests).

Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
FranzBusch: That is highly unfortunate since it should be synchronously present. Let's file an issue to track this since in the `HTTPAPIs` we should not make the property `async` or `throws`

FranzBusch (approved):

Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
gjcairo: Why should it be synchronously present and not throw?

This is currently async because we're are calling `get()` on an ELF. We could just block if we need it to be sync. 

We can't get around the `throws` unless we `try?`, but is hiding the errors desired?

Sources/NIOHTTPServer/NIOHTTPServerDefaultConnectionHandler.swift
gjcairo: I prefer to keep this. Even though it's optional in Swift to have conformances declare `throws` if they don't throw, if the conformances are public, it would be breaking to eventually make them throwing, even if the protocol originally declared it as `throws`.

Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
FranzBusch: Because at the point where we hand the connection to the user we already have the TLS connection established. So we should be able to get the peer cert before we hand the connection to the user

aryan-25 (approved):

Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
gjcairo: Sure. We can probably get it via sync operations. I'll create a follow up issue.

gjcairo: https://github.com/swift-server/swift-http-server/issues/97

PR #94, opened 2026-07-06
Add `CloseableConnection` capability

gjcairo:
This is a follow-up of the connection lifecycle PR: https://github.com/swift-server/swift-http-server/pull/92

It adds support for signalling that the current connection should close after the in-flight response via a new `CloseableConnection` `HTTPServerCapability`. 
`NIOHTTPServer.RequestContext` conforms to it and exposes this capability via `requestContext.signalConnectionClose()`.


Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
aryan-25: Why is the default value `true`?

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
```diff
@@ -146,20 +178,21 @@ final class HTTPKeepAliveHandler: ChannelDuplexHandler {
     func flush(context: ChannelHandlerContext) {
         // An upstream writer forced a flush. Same deadline as `channelReadComplete`:
         // release any buffered parts, with `Connection: close` if request `.end`
-        // hasn't arrived.
+        // hasn't arrived or close was signalled.
         if case .buffering = self.finalResponseState {
             self.flushBuffer(context: context)
         }
         context.flush()
```
aryan-25: If `self.flushBuffer()` was called above, then `context.flush()` will be called twice (`self.flushBuffer()` also calls `context.flush()`).

Tests/NIOHTTPServerTests/Utilities/TestingChannelClientServer/NIOHTTPServer+HTTP1.swift
```diff
@@ -33,7 +33,7 @@ extension NIOHTTPServer {
     {
         // The server requires a NIOAsyncChannel, so we create one from the test channel
         let serverTestAsyncChannel = try await testChannel.eventLoop.submit {
-            try NIOAsyncChannel<NIOAsyncChannel<HTTPRequestPart, HTTPResponsePart>, Never>(
+            try NIOAsyncChannel<NIOHTTPServer.HTTP1ChildConnection, Never>(
```
aryan-25: nit:
```suggestion
            try NIOAsyncChannel<HTTP1ChildConnection, Never>(
```

Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
```diff
@@ -163,37 +173,51 @@ extension NIOHTTPServer {
...
 
     /// Builds a ``ConnectionContext`` for an HTTP/1.1 request channel.
+    ///
+    /// The context's ``ConnectionContext/signalConnectionClose()`` synchronously
+    /// flips the shared close flag the channel's ``HTTPKeepAliveHandler``
+    /// observes when writing the next response head. The synchronous set
+    /// side-steps any race between firing a NIO event off-loop and writing the
+    /// response head off-loop. The handler reacts by amending the next response
+    /// head with `Connection: close` and closing the channel once the response
+    /// `.end` is written.
```
aryan-25: nit: I think this detail is unnecessary for this function.

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
```diff
@@ -374,6 +383,13 @@ extension NIOHTTPServer {
     }
 
     /// Handles an HTTP/2 stream channel, which carries exactly one request per stream.
+    ///
+    /// If the request handler invokes
+    /// ``ConnectionContext/signalConnectionClose()`` while running, the
+    /// connection-context's `closeAction` fires `ChannelShouldQuiesceEvent` on
+    /// the connection-channel pipeline; NIO's HTTP/2 connection management
+    /// handler reacts by sending `GOAWAY`, letting other in-flight streams
+    /// complete normally, and finally closing the connection.
```
aryan-25: nit: I think this detail is unnecessary for this function.

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
```diff
@@ -146,6 +147,13 @@ extension NIOHTTPServer {
     }
 
     /// Builds a ``ConnectionContext`` for an HTTP/2 connection channel.
+    ///
+    /// The context's ``ConnectionContext/signalConnectionClose()`` fires
+    /// `ChannelShouldQuiesceEvent` on the connection-channel pipeline. NIO's
+    /// `NIOHTTP2ServerConnectionManagementHandler` (added to that pipeline by
+    /// `configureAsyncHTTP2Pipeline`) reacts by initiating graceful shutdown —
+    /// sending `GOAWAY`, letting in-flight streams complete normally, and
+    /// finally closing the connection.
```
aryan-25: nit: I think this detail is unnecessary for this function.

Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
```diff
@@ -163,37 +173,51 @@ extension NIOHTTPServer {
...
                 .syncOperations
                 .addTimeoutHandlers(self.configuration.connectionTimeouts)
 
-            return try NIOAsyncChannel<HTTPRequestPart, HTTPResponsePart>(
+            let asyncChannel = try NIOAsyncChannel<HTTPRequestPart, HTTPResponsePart>(
                 wrappingChannelSynchronously: channel,
                 configuration: asyncChannelConfiguration
             )
+            return HTTP1ChildConnection(asyncChannel: asyncChannel, closeFlag: closeFlag)
         }
```
aryan-25: We should initialise `closeFlag` inside the `flatMapThrowing` closure.

Tests/NIOHTTPServerTests/ConnectionLifecycleTests.swift
```diff
@@ -293,6 +293,291 @@ struct ConnectionLifecycleTests {
...
+        let numInflight = 3
+        let allRequestsReceived = elg.any().makePromise(of: Void.self)
+        let arrivedCounter = NIOLockedValueBox(0)
+
+        let connectionHandler = NIOHTTPServerDefaultConnectionHandler(
+            handler: HTTPServerClosureRequestHandler {
+                request,
+                requestContext,
+                reader,
+                responseSender in
```
aryan-25: formatting nit: it should be possible to put all these arguments on the same line. `request` and `reader` can be replaced with `_` if the line length becomes too long. The same applies for all the other test cases below.

Tests/NIOHTTPServerTests/ConnectionLifecycleTests.swift
```diff
@@ -293,6 +293,291 @@ struct ConnectionLifecycleTests {
...
+                let head = try await iterator.next()
+                guard case .head(let response) = head else {
+                    Issue.record("Expected response head but got \(String(describing: head))")
+                    return
+                }
+                #expect(response.headerFields[.connection] == "close")
+                // Look for any duplicate `Connection` headers; the field should
+                // appear exactly once with value "close".
+                let connectionValues = response.headerFields[values: .connection]
+                #expect(connectionValues == ["close"])
```
aryan-25:
```suggestion
                // Look for any duplicate `Connection` headers; the field should
                // appear exactly once with value "close".
                let connectionValues = response.headerFields[values: .connection]
                #expect(connectionValues == ["close"])
```

Tests/NIOHTTPServerTests/ConnectionLifecycleTests.swift
```diff
@@ -293,6 +293,291 @@ struct ConnectionLifecycleTests {
...
+
+                while let part = try await iterator.next() {
+                    if case .end = part { break }
+                }
+                let _trailing = try await iterator.next()
+                #expect(_trailing == nil)
```
aryan-25: nit: the leading underscore in the variable name seems unnecessary. The same applies for the test case below.

Sources/NIOHTTPServer/NIOHTTPServer+ConnectionContext.swift
```diff
@@ -74,5 +92,22 @@ extension NIOHTTPServer {
                 return nil
             }
         }
+
+        /// Signal that the connection should close after the current response.
+        ///
+        /// Non-blocking and idempotent. Effective after the current response:
+        ///
+        /// - On HTTP/1.1, the response carries `Connection: close` and the
+        ///   channel is closed once the response has been written.
```
aryan-25: We should document that `signalConnectionClose()` must be called *before* sending the response for this to hold true.

Tests/NIOHTTPServerTests/ConnectionLifecycleTests.swift
```diff
@@ -293,6 +293,291 @@ struct ConnectionLifecycleTests {
...
+                    return
+                }
+                #expect(response.status == .ok)
+                // We deliberately don't assert anything about the `Connection`
+                // header here: `responseSender.send(_:)` returns once the head
+                // is yielded to the async-stream queue, which may be before the
+                // keep-alive handler has actually processed it on the event
+                // loop. So the flag set immediately after `send` is sometimes
+                // still observed in time to amend the head. The reliable
+                // observable is that the channel closes after the response.
```
aryan-25: We can bypass this by creating a promise that the client will fulfil after it receives the head. The server should await that promise before calling `requestContext.signalConnectionClose()`.

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
gjcairo: Because it needs to be initialised to something and the invariant for latter iterations is that it should be true in between iterations. In reality it doesn't really make much of a difference for the first request because we set it to false when we receive a request head anyways. However this was confusing so I've renamed this to `requestBodyInFlight` and flipped the semantics, which is a lot clearer.

Sources/NIOHTTPServer/HTTPKeepAliveHandler.swift
gjcairo: I've tidied this up

Sources/NIOHTTPServer/NIOHTTPServer+ConnectionContext.swift
gjcairo: Unless I'm misunderstanding what you're saying, this isn't true. We can call signalConnectionClose() mid-response and that's okay, we will close once response has been written.

aryan-25: I was talking about initiating connection close _after_ the response header has been flushed.

I think this doc should be updated with these two things:
- The `Connection: close` header cannot be written if the response head was already flushed before `signalConnectionClose()` being called.
- In this case though, the channel will be closed after the end part is written (following from your change in commit [d0688c1](https://github.com/swift-server/swift-http-server/pull/94/commits/d0688c1b5815aefd57710196bb7f9f08fc0dbbf3)).

Tests/NIOHTTPServerTests/ConnectionLifecycleTests.swift
```diff
@@ -293,6 +289,348 @@ struct ConnectionLifecycleTests {
...
+
+        let connectionHandler = NIOHTTPServerDefaultConnectionHandler(
+            handler: HTTPServerClosureRequestHandler { _, requestContext, reader, responseSender in
+                var reader = reader
+                // Drain the body so `invokeHandler` recovers the iterator and the
+                // request loop stays alive — otherwise the loop exits when the
+                // handler returns, and the dispatcher closes the channel anyway.
+                var body = UniqueArray<UInt8>()
+                body.reserveCapacity(1024)
+                _ = try await reader.collect(into: &body)
```
aryan-25, line 532: This test seems very similar to the one above. These lines are the only difference I see compared to the test above. But I don't think this should matter because the client in this test case only sends a `head` and `end` part.

gjcairo: It does matter because of the way we handle the `ReaderState` internally when we finish reading vs when we don't (see `NIOHTTPServer/invokeHandler`). Also, another difference is we're not setting any timeouts in this test: we want to make sure the connection is being closed and not hang indefinitely/be closed by other timeout mechanisms instead of close being signalled.

Tests/NIOHTTPServerTests/ConnectionLifecycleTests.swift
```diff
@@ -293,6 +289,348 @@ struct ConnectionLifecycleTests {
...
+            transportSecurity: .plaintext
+        )
+        configuration.connectionTimeouts = .init(idle: nil, readHeader: nil, readBody: nil)
+        let server = NIOHTTPServer(logger: Self.serverLogger, configuration: configuration)
+
+        let (canSignalStream, canSignal) = AsyncStream<Void>.makeStream()
+
+        let connectionHandler = NIOHTTPServerDefaultConnectionHandler(
+            handler: HTTPServerClosureRequestHandler { _, requestContext, reader, responseSender in
+                var reader = reader
```
aryan-25: `reader` doesn't get mutated; you can remove this line.

aryan-25 (approved):
Thank you! Just one small comment.

Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
```diff
@@ -41,4 +41,21 @@ extension HTTPServerCapability {
...
+    /// A request-context capability that lets a request handler signal that the
+    /// connection should close after the current response.
+    ///
+    /// Servers whose request context conforms to this capability allow handlers
+    /// to indicate that the underlying connection should be closed once the
+    /// in-flight response has been sent. Implementations make the signal
+    /// effective on a best-effort basis (HTTP/1.1 typically appends
+    /// `Connection: close` and closes the channel; HTTP/2 typically sends
+    /// `GOAWAY` and lets in-flight streams complete normally).
+    public protocol CloseableConnection: RequestContext {
```
FranzBusch, line 54: I find the semantics a bit weird that this only closes after the response is send. Why doesn't it close right away?

gjcairo: For H1, closing after the response is sent allows users to write back a response that will contain the `Connection:close` header, because we'll set it when signalling close.

For H2, I think it's weird to just abruptly close the connection because there can be other streams open. We currently send GOAWAY but I suppose we could send `RST_STREAM`s instead?

I think just abruptly closing the connection feels too harsh, but perhaps that's the desired behaviour.

cc @ehaydenr

ehaydenr: I think you want to be able to do both. If you're shutting down, you may first want to send a graceful GOAWAY to avoid disrupting in flights requests. After some grace period, you may follow up with a more abrupt GOAWAY followed by connection close which would break streams. Alternatively, you may want to immediately close things abruptly if the peer is behaving badly.

gjcairo: You think it would make sense to be able to initiate both from here, or would you have separate capabilities? 

I've not given this much thought but one thing that kinda bothers me is that force closing doesn't stop you from still attempting to read or write within this handler, and that will cause errors. Basically the only thing you can do after a force close is to return. The only way I can think of doing this is by having the force close consume the reader and response sender/writer, but I don't think that's particularly nice either.

ehaydenr: I think if you force close and then do a read or write, it should probably result in an error and that's OK

FranzBusch: @ehaydenr do you wanna be able to do both the graceful shutdown and the force closure from a stream? I am actually wondering if we are doing this the wrong way around. In my opinion, having this as a stream capability is really weird and it would be better served as API on [Connection](https://github.com/swift-server/swift-http-server/blob/0dc060b94856e54be07cc28e488738a2e3e78e49/Sources/NIOHTTPServer/NIOHTTPServer%2BConnection.swift#L35). So what I am suggesting is that `Connection` gains methods for `triggerGracefulShutdown`, `close` and maybe a combined one that triggers graceful shutdown and after a duration closes. You can then inject an `MPSC` channel into your stream handlers that will do this on the connection. What do you think about this?

ehaydenr: I want to be able to do it from a request hander. Whether it's exposed as a RequestContext capability or I need to plumb it through to where I instantiate my request handler, I don't think it matters. Exposing as a capability seems like the natural way to accomplish this. What about it do you find weird?

FranzBusch: I continue to think it is weird to expose connection lifecycle management methods on request context. I feel like we will end up with a god request that will allow you to do everything with the connection. I much prefer clear separation of concerns between the request handler and the mostly read-only context of the connect and a separate connection lifecycle API. If users want to tie the two together they can setup a communication channel between those.

PR #96, opened 2026-07-13
Add unix domain socket support for bind targets

### LINKED ISSUE #69: Support binding to a unix socket
Currently only IP address and port is supported as a bind target, unix socket binding is missing

mob-connection:
**Motivation**

NIOHTTPServer could only bind to a host and port. Serving over a UNIX domain socket is a common requirement (e.g. a reverse proxy talking to the server over a local socket), so this PR adds a UDS bind target.

**Modifications**

• Added BindTarget.unixDomainSocket(path:) (typed as System.FilePath) and a matching SocketAddress case; the path is converted to String only at the NIO edge (bind(unixDomainSocketPath:) and fileIO.unlink).
• Bind over a UNIX domain socket for both plaintext (HTTP/1.1) and secure-upgrade (HTTP/1.1 + HTTP/2) channels, and report the bound path from listeningAddresses.
• Remove the socket file on shutdown; fail the bind if the path is already occupied so a stale socket is never silently reused.
• Support a socketPath key in swift-configuration, mutually exclusive with host/port.
• Reject UNIX domain socket bind targets when HTTP/3 is enabled (HTTP/3 runs over QUIC/UDP), with a dedicated configuration error.
• Added tests for the listening-address mapping and the HTTP/3 rejection.
• Factored the bind-target dispatch into a ServerBootstrap.bind(_:to:childChannelInitializer:) helper so the host/port and UDS paths aren't duplicated across the HTTP/1.1 and secure-upgrade setup.

**Result**

NIOHTTPServer can now listen on a UNIX domain socket over HTTP/1.1 and HTTP/2. Resolves [#69](https://github.com/swift-server/swift-http-server/issues/69).


mob-connection:
For #69 I used a struct-backed `NIOHTTPServer.SocketAddress` with an added
`.unixDomainSocket(path:)` case, so it stays extensible (e.g. vsock later)
without a source-breaking change.

Do you think it makes sense to expose it as a plain `enum` instead ?

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
```diff
@@ -47,8 +48,22 @@ public struct NIOHTTPServerConfiguration: Sendable {
...
+        /// Creates a bind target for a unix domain socket.
+        ///
+        /// - Parameter path: The file system path to bind the unix domain socket to (e.g., "/tmp/server.sock")
+        /// - Returns: A configured `BindTarget` instance
+        ///
+        /// ## Example
+        /// ```swift
+        /// let target = BindTarget.unixDomainSocket(path: "/tmp/server.sock")
+        /// ```
+        public static func unixDomainSocket(path: String) -> Self {
```
FranzBusch: This should probably use the new `FilePath` type once it becomes available

mob-connection: > This should probably use the new `FilePath` type once it becomes available

In the UDS bind support the socket path has two different types depending on direction: on input we take a FilePath (BindTarget.unixDomainSocket(path:)), but on output we expose it as String (SocketAddress.unixDomainSocketPath, mirroring NIOCore.SocketAddress.pathname). If we commit to FilePath, I'd lean toward using it on output too for consistency — but keeping String also makes sense since that's just what NIO reports back. Which way should we go?

0xTim: Yeah I think ignore what NIO does and keep the API what looks the best in Swift, in this case `FilePath`s for all types that represent instead of a String

0xTim (commented):
Looking mostly good, I think we should add H2 support as well though

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
```diff
@@ -344,6 +360,17 @@ public struct NIOHTTPServerConfiguration: Sendable {
...
+        // HTTP/3 runs over QUIC/UDP and cannot be served over a unix domain socket.
+        if supportedHTTPVersions.http3ConfigIfSupported != nil {
+            for bindTarget in bindTargets {
+                if case .unixDomainSocket = bindTarget.backing {
+                    throw NIOHTTPServerConfigurationError.unixDomainSocketNotSupportedOverHTTP3
+                }
+            }
+        }
+        #endif
+
```
0xTim, line 431: What about HTTP/2?

mob-connection: > What about HTTP/2?

HTTP/2 over a UDS works today: it's negotiated via ALPN on the secure-upgrade path, which already handles .unixDomainSocket binding, so no special-casing is needed (unlike HTTP/3, which is QUIC/UDP).

Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
```diff
@@ -148,6 +148,30 @@ extension NIOHTTPServer {
                         )
                     }
                     serverChannels.append((serverChannel, serverQuiescingHelper))
+                case .unixDomainSocket(let path):
```
aryan-25: To avoid duplication with the case above, we should define an extension on `ServerBootstrap`'s `bind` method that accepts a `NIOHTTPServerConfiguration.BindTarget`. That method should switch over the bind target and call into either [the host and port variant](https://github.com/apple/swift-nio/blob/48119dbbd23e3eabba48952ac7f75ebeeb87c217/Sources/NIOPosix/Bootstrap.swift#L538) or [the UDS variant](https://github.com/apple/swift-nio/blob/48119dbbd23e3eabba48952ac7f75ebeeb87c217/Sources/NIOPosix/Bootstrap.swift#L597) of `ServerBootstrap.bind`.

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
```diff
@@ -247,6 +247,27 @@ extension NIOHTTPServer {
                         )
                     }
                     serverChannels.append((serverChannel, serverQuiescingHelper))
+                case .unixDomainSocket(let path):
```
aryan-25: Same as above. You will be able to use the extension here too.

aryan-25 (changes requested):

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -164,6 +165,10 @@ public struct NIOHTTPServer: HTTPServer {
 
         let serverChannels = try await self.makeServerChannels()
 
+        // Remove the socket files for any UDS bind targets so their paths are freed for the next run.
+        // Registered only after all binds succeeded, so every path is one we created.
+        defer { await self.removeUNIXDomainSocketFiles() }
```
mob-connection, line 170: serve(...) now has two defer blocks — `finishListeningAddressPromise()` and `await removeUNIXDomainSocketFiles()`. Just flagging that for visibility in case it's worth consolidating or reordering; happy to adjust if you'd prefer a different structure. 🙌

mob-connection:
Deduplicated the bind logic (@aryan-25's comments): added a `ServerBootstrap.bind(_:to:childChannelInitializer:)` helper that switches over the BindTarget and dispatches to the host/port or UDS overload, so `setupHTTP1_1ServerChannels` and `setupSecureUpgradeServerChannels` no longer duplicate the setup. One note: it's a static method taking sending ServerBootstrap — under strict concurrency self.bind on the non-Sendable bootstrap trips "sending 'self' risks data races", and self can't be marked sending.
   • Merged latest main and resolved the conflicts; all tests pass locally.

Left one small note inline about the two defer blocks in serve(...). Happy to iterate further — thanks again! 🙌

PR #98, opened 2026-07-14
Introduce HTTP/3 configuration

aryan-25:
### Motivation:

[`swift-nio-quic`](https://github.com/apple/swift-nio-quic) and [`swift-nio-http3`](https://github.com/apple/swift-nio-http3) have recently been announced, paving the way for HTTP/3 support in `NIOHTTPServer`.

This PR introduces HTTP/3 configuration types as a first step towards HTTP/3 support.

### Modifications:

- Introduced a new `HTTP3` type under `NIOHTTPServerConfiguration` containing three sub-components, namely `QUICConfiguration`, `ProtocolConfiguration`, and `ConnectionSettings`. These types wrap over the corresponding `NIOQUIC` and `NIOHTTP3` types.

- Added `swift-nio-quic`, `swift-nio-quic-helpers`, and `swift-nio-http3` dependencies.

### Result:

HTTP/3 configuration types are now defined.


.github/workflows/apibreakage.yml
gjcairo, line 1: Can't we amend the API breakage step like we did for the other CI steps in the main/PR YAMLs?

Sources/NIOHTTPServer/Configuration/HTTP3/HTTP3+ConnectionSettings.swift
```diff
@@ -0,0 +1,60 @@
...
+        public var qpackBlockedStreams: UInt64
+
+        /// The maximum size of a field section. The default value is `nil`, which is interpreted as there being no
+        /// field section size limit.
+        ///
+        /// - SeeAlso: https://www.rfc-editor.org/rfc/rfc9114.html#section-7.2.4.1-2.2.1. Corresponds to
+        ///   `SETTINGS_MAX_FIELD_SECTION_SIZE`.
+        public var maximumFieldSectionSize: UInt64?
+
+        /// The default configuration.
```
gjcairo: I'm thinking we should perhaps move the comments about default values to here instead of having them at the property level, since they're really set at this point. This applies to all types.

gjcairo (changes requested):

.github/workflows/apibreakage.yml
aryan-25: Unfortunately not, since the [soundness.yml](https://github.com/swiftlang/github-workflows/blob/main/.github/workflows/soundness.yml) in `swiftlang/github-workflows` does not expose a config for specifying environment variables.

This is what we had to do in [swift-nio-http3](https://github.com/apple/swift-nio-http3/pull/7) and [swift-nio-quic](https://github.com/apple/swift-nio-quic/pull/38) too.

gjcairo: Ah, that's unfortunate.

Sources/NIOHTTPServer/Configuration/HTTP3/HTTP3+ConnectionSettings.swift
aryan-25: Addressed in commit 57f8229.

gjcairo: Thanks. With "all types" I meant other files as well, since we have this pattern elsewhere too. Would you mind making that change?

aryan-25: Addressed in commit 555877d.

.github/workflows/apibreakage.yml
FranzBusch, line 1: We shouldn't need this but instead change the upstream swiftlang workflow to run the prebuild commands at the right place so we can set the env variable

Sources/NIOHTTPServer/Configuration/HTTP3/NIOHTTPServerConfiguration+HTTP3.swift
```diff
@@ -0,0 +1,66 @@
...
+
+@available(anyAppleOS 26.0, *)
+extension NIOHTTPServerConfiguration {
+    /// Configuration for HTTP/3.
+    public struct HTTP3: Sendable, Hashable {
+        /// QUIC transport configuration.
+        public var quicConfiguration: QUICConfiguration = .defaults
+
+        /// HTTP/3 protocol-level configuration.
+        public var protocolConfiguration: ProtocolConfiguration = .defaults
```
FranzBusch: Shouldn't this be not nested? I find the naming protocol configuration weird? Maybe we just put them into this level of config without nesting

FranzBusch (commented):
We should add support for `swift-configuration` for all those new options

.github/workflows/apibreakage.yml
josephnoir: I hadn't opened the PR for that yet. Now open here: https://github.com/swiftlang/github-workflows/pull/304

Sources/NIOHTTPServer/Configuration/HTTP3/NIOHTTPServerConfiguration+HTTP3.swift
aryan-25: Yeah, we also have the protocol-level configs for HTTP/2 in the same config type (not nested).
I've fixed this in commit ffdfded.

aryan-25:
I was planning to add `swift-configuration` integration in a follow-up PR in order to keep the diff minimal in this PR.

.github/workflows/apibreakage.yml
aryan-25: Thank you @josephnoir!

I've now updated `pull_request.yml` in 40dc5f7 to use your change. Note that since the change hasn't been released, I've pinned the soundness workflow to your commit.

FranzBusch (commented):
I think the public API now looks good. Can we introduce all of this behind a new `HTTP3` trait that is off by default. This is mostly since the other libraries are not yet stable and will allow us to break more easily.

FranzBusch (approved):

PR #101, opened 2026-07-20
Add support for serving HTTP/3

aryan-25:
### Motivation

Following from #98, this PR introduces support for serving HTTP/3.

### Modifications

- Added `NIOHTTPServer+HTTP3.swift` which contains helper methods for setting up a QUIC channel, adding HTTP/3 handlers, and serving HTTP/3 connections.
- Wired HTTP/3 into `makeServerChannels()`.
  - When HTTP/3 is the only version specified in `supportedHTTPVersions`, only a QUIC channel is bound to the specified port. When `supportedHTTPVersions` includes HTTP/1.1 and/or HTTP/2 alongside HTTP/3, the QUIC UDP channel is bound on the same port as the TCP channel serving HTTP/1.1 and/or HTTP/2.
  - Note: I will introduce Alt-Svc advertising in a subsequent PR.

### Result

`NIOHTTPServer` can now serve HTTP/3.

### DIFF
--- a/Sources/NIOHTTPServer/NIOHTTPServer+HTTP3.swift
+++ b/Sources/NIOHTTPServer/NIOHTTPServer+HTTP3.swift
@@ -1,0 +1,276 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift HTTP Server open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift HTTP Server project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift HTTP Server project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if HTTP3
+import HTTP3
+import Logging
+import NIOCore
+import NIOEmbedded
+@_spi(HTTP3AsyncInterface) import NIOHTTP3
+import NIOHTTPTypes
+import NIOPosix
+import NIOQUIC
+import NIOQUICHelpers
+import NIOSSL
+import X509
+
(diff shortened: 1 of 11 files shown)


Sources/NIOHTTPServer/NIOHTTPServer+HTTP3.swift
```diff
@@ -0,0 +1,276 @@
...
+        _ http3Connection: HTTP3ServerConnection<
+            NIOAsyncChannel<HTTPRequestPart, HTTPResponsePart>,
+            NIOQUIC.QUICStreamCreator
+        >,
+        handler: Handler
+    ) async {
+        let context = ConnectionContext(
+            httpVersion: .http3,
+            remoteAddress: nil,
+            localAddress: nil,
```
FranzBusch, line 62: What about those?

FranzBusch (approved):

Sources/NIOHTTPServer/NIOHTTPServer+HTTP3.swift
aryan-25: [`HTTP3ServerConnectionMultiplexer`](https://github.com/apple/swift-nio-http3/blob/240d833a3bf6cfb547e034861792377ac0fa23eb/Sources/NIOHTTP3/HTTP3ConnectionMultiplexer.swift#L170) vends out a [`HTTP3ServerConnection`](https://github.com/apple/swift-nio-http3/blob/240d833a3bf6cfb547e034861792377ac0fa23eb/Sources/NIOHTTP3/HTTP3ConnectionMultiplexer.swift#L25) for each connection.

The problem is that the underlying connection channel cannot be accessed from `HTTP3ServerConnection` in its current shape.

I've set these properties to `nil` until we make some changes in `HTTP3ServerConnection`.

PR #104, opened 2026-07-29
Add reset stream API

gjcairo:
Until now, a request handler had no way to abort the stream carrying a request.

This PR adds a reset operation to the response side:
- `ResponseSender.reset(_:)`: abort before any response head is sent.
- `Writer.reset(_:)`: abort a response that has already started.

Because a reset is inherently protocol-specific, reset hands the handler a protocol-discriminated `NIOHTTPServer.StreamReset` to switch over:
  - `.http2` → sends an HTTP/2 RST_STREAM with a chosen `HTTP2ErrorCode`.
  - `.http3` → sends an HTTP/3 RESET_STREAM with a chosen QUIC application error code.
  - `.unavailable` → HTTP/1.1, which has no per-stream reset.

`reset` consumes the sender/writer, so users are prevented from writing anything further once the stream has been reset.

An example scenario using `CONNECT`:
```swift
  let writer = try await responseSender.send(.init(status: .ok))   // e.g. CONNECT tunnel established
  // … the tunnelled connection fails …
  try writer.reset { streamReset in
      switch consume streamReset {
      case .http2(let h2): 
        h2.reset(code: .connectError)
      case .http3(let h3): 
        h3.reset(code: 0x010f)   // H3_CONNECT_ERROR
      case .unavailable:
        break
      }
  }
```

### DIFF
--- a/Sources/NIOHTTPServer/NIOHTTPServer+StreamReset.swift
+++ b/Sources/NIOHTTPServer/NIOHTTPServer+StreamReset.swift
@@ -1,0 +1,100 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift HTTP Server open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift HTTP Server project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift HTTP Server project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+public import NIOHTTP2
+import NIOHTTPTypes
+import NIOHTTPTypesHTTP2
+
+#if HTTP3
+import NIOQUICHelpers
+#endif  // HTTP3
+
+@available(anyAppleOS 26.0, *)
+extension NIOHTTPServer {
+    /// The protocol-specific surface for resetting the stream carrying a request.
+    ///
+    /// Stream resets only exists over HTTP/2 and HTTP/3. The only abrupt tear-down mechanism available
+    /// for HTTP/1.1 is closing the connection.
+    @nonexhaustive
+    public enum StreamReset: ~Copyable {
+        /// The protocol has no per-stream coded reset (for example HTTP/1.1).
+        ///
+        /// There is nothing to reset with a code here. Returning from the handler without concluding the response
+        /// aborts the exchange and the connection is closed.
+        case unavailable
+
+        /// The connection is HTTP/2; ``HTTP2StreamReset`` sends a `RST_STREAM`.
+        case http2(HTTP2StreamReset)
+
+        #if HTTP3
+        /// The connection is HTTP/3; ``HTTP3StreamReset`` sends a QUIC `RESET_STREAM`.
+        case http3(HTTP3StreamReset)
+        #endif  // HTTP3
+    }
+
+    /// Resets an HTTP/2 stream by sending a `RST_STREAM` frame with a chosen error code.
+    public struct HTTP2StreamReset: ~Copyable {
+        private let channel: any Channel
+
+        init(channel: any Channel) {
+            self.channel = channel
+        }
+
+        /// Sends a `RST_STREAM` frame for this stream with the provided error code.
+        ///
+        /// - Parameter code: The `RST_STREAM` error code to send.
+        public consuming func reset(code: HTTP2ErrorCode) {
+            // The `HTTP2FramePayloadToHTTPServerCodec` on the stream channel translates this event into an `RST_STREAM`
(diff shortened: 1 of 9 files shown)


ehaydenr:
Related https://github.com/apple/swift-nio-http3/pull/14

Sources/NIOHTTPServer/NIOHTTPServer+StreamReset.swift
```diff
@@ -0,0 +1,100 @@
...
+
+        init(channel: any Channel) {
+            self.channel = channel
+        }
+
+        /// Sends a `RST_STREAM` frame for this stream with the provided error code.
+        ///
+        /// - Parameter code: The `RST_STREAM` error code to send.
+        public consuming func reset(code: HTTP2ErrorCode) {
+            // The `HTTP2FramePayloadToHTTPServerCodec` on the stream channel translates this event into an `RST_STREAM`
```
aryan-25, line 59: nit:
```suggestion
            // The `HTTP2FramePayloadToHTTPServerCodec` on the stream channel translates this event into a `RST_STREAM`
```

aryan-25 (approved):

Sources/NIOHTTPServer/NIOHTTPServer+StreamReset.swift
```diff
@@ -0,0 +1,100 @@
...
+#endif  // HTTP3
+
+@available(anyAppleOS 26.0, *)
+extension NIOHTTPServer {
+    /// The protocol-specific surface for resetting the stream carrying a request.
+    ///
+    /// Stream resets only exists over HTTP/2 and HTTP/3. The only abrupt tear-down mechanism available
+    /// for HTTP/1.1 is closing the connection.
+    @nonexhaustive
+    public enum StreamReset: ~Copyable {
```
FranzBusch, line 31: I am not very happy that this surfaces different APIs for H/2 and H/3. In the end both are just wrappers around `Int` which is treated as an application error code right? Do we need to surface this difference really?

Sources/NIOHTTPServer/NIOHTTPServer+StreamReset.swift
```diff
@@ -0,0 +1,100 @@
...
+        private let channel: any Channel
+
+        init(channel: any Channel) {
+            self.channel = channel
+        }
+
+        /// Sends a `RST_STREAM` frame for this stream with the provided error code.
+        ///
+        /// - Parameter code: The `RST_STREAM` error code to send.
+        public consuming func reset(code: HTTP2ErrorCode) {
```
FranzBusch, line 58: So this method executes I/O and it must be `async` and `throws`.

Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
```diff
@@ -30,7 +31,23 @@ extension NIOHTTPServer {
...
+        }
+
+        /// Abandons the response and resets the stream carrying this request.
+        ///
+        /// Call this instead of ``send(_:)`` when the request should be aborted before any response head is sent. This
+        /// consumes the sender, so no response can be sent afterwards.
+        ///
+        /// - Parameter body: A closure that is provided a ``NIOHTTPServer/StreamReset`` instance from which the request
+        ///   stream can be reset with a transport-specific error code.
+        public consuming func reset(_ body: (consuming NIOHTTPServer.StreamReset) throws -> Void) throws {
```
FranzBusch, line 48: I like that this is a method on the response sender but I am not sure we need the closure yet. Wouldn't it suffice if we just made this take an `Int` as the application error code? It probably should also consume the reader otherwise the reader is left in a bad state.

Sources/NIOHTTPServer/NIOHTTPServer+StreamReset.swift
gjcairo: There were two reasons why I decided to go this way, but basically it boils down to request handlers potentially wanting to do different things depending on the HTTP version they're running on (to fail with different error codes for example). The first reason is because the errors that can be surfaced aren't the same for H2 and H3; I considered having a set of common errors that could be mapped across each version, but not many of them actually overlap, and having some "raw" code where most of the errors would have to live felt weird.
The second and perhaps more importantly is that H1 does not support resets at all, so handlers may want to do close the connection instead or do something different altogether.

Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
gjcairo: We could potentially return the `StreamReset` instead of having the closure, yeah.

As to consuming the reader, I played with that, but for H3 where only the sending direction is closed, we may still have bytes to read.

gjcairo:
Closed in favour of https://github.com/swift-server/swift-http-server/pull/115

PR #105, opened 2026-07-30
Improve test coverage

aryan-25:
Motivation:

All tests currently only test the plaintext or secure upgrade transport. It would be useful to also include the HTTP/3 transport in these tests. Due to the differences in setting up a secure upgrade client compared to an HTTP/3 client, this requires some refactoring to the test utilities.

Modifications:

- Moved all test helpers from `NIOHTTPServerTests` to a new `TestHelpers` type to make accessing the helpers more convenient.
- Renamed `NegotiatedClientConnection` to `TestClientConnection` and extended it to also wrap HTTP/3 connections.
- Introduced convenience helpers for:
  - Setting up a client connection, or a request channel directly, through with-style helper methods.
  - Creating a server and a matching client configuration for a given transport.
- Added a test HTTP/3 client.
- Parameterised tests over `HTTPVersion` and removed some tests that became redundant as a result.

Result:

Better test coverage.


Tests/NIOHTTPServerTests/Utilities/NIOClient/NIOClient+HTTP3.swift
```diff
@@ -0,0 +1,156 @@
...
+
+    func createNewConnection(
+        serverName: String,
+        remoteAddress: SocketAddress,
+        connectionInitializer h3ConnectionInitializer: @escaping @Sendable (any Channel) -> EventLoopFuture<Void>
+    ) -> EventLoopFuture<any Channel> {
+        self.quicHandler.createOutboundConnection(
+            serverName: serverName,
+            remoteAddress: remoteAddress,
+            connectionInitializer: { [connectionInitializer] connectionChannel, streamCreator in
```
gjcairo, line 41: Nit: I don't think this capture is needed, it will be implicitly captured anyways
```suggestion
            connectionInitializer: { connectionChannel, streamCreator in
```

Tests/NIOHTTPServerTests/Utilities/HTTPVersion.swift
```diff
@@ -12,19 +12,46 @@
 //
 //===----------------------------------------------------------------------===//
 
-enum HTTPVersion {
-    case http1_1
-    case http2
+import NIOHTTPServer
 
+@available(anyAppleOS 26.0, *)
+extension NIOHTTPServer.HTTPVersion {
```
gjcairo, line 18: Don't we now have this type in the main package?

Tests/NIOHTTPServerTests/Utilities/TestClientConnection.swift
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+
+    /// Closes the underlying connection.
+    func close() async throws {
+        switch self.connectionProtocol {
+        case .http1(let asyncChannel):
+            do {
+                try await asyncChannel.channel.close()
+            } catch ChannelError.alreadyClosed {
```
gjcairo, line 91: Is it okay to swallow this error everywhere? why are we doing this?

Tests/NIOHTTPServerTests/NIOHTTPServer+ServiceLifecycleTests.swift
```diff
@@ -25,24 +27,29 @@ import Testing
...
-    let serverLogger = Logger(label: "NIOHTTPServiceLifecycleTests")
-    let serviceGroupLogger = Logger(label: "NIOHTTPServiceLifecycleTests_ServiceGroup")
+    let clientLogger = Logger(label: "NIOHTTPServiceLifecycleTests.client")
+    let serverLogger = Logger(label: "NIOHTTPServiceLifecycleTests.server")
+    let serviceGroupLogger = Logger(label: "NIOHTTPServiceLifecycleTests.serviceGroup")
 
     @Test(
         "Active connection completes when graceful shutdown triggered",
-        arguments: [HTTPVersion.http1_1, HTTPVersion.http2]
+        arguments: [NIOHTTPServer.HTTPVersion.http1_1, .http2]
```
gjcairo, line 44: We should make sure we're exercising H3 in all tests that make sense, like this one. It's fine to have a follow up for this.

Tests/NIOHTTPServerTests/Utilities/TestClientConnection.swift
```diff
@@ -0,0 +1,237 @@
...
+                    connectionProtocol: .http3(connection: h3Connection, quicChannel: quicChannel)
+                )
+            } catch {
+                try? await quicChannel.close()
+                throw error
+            }
+        #endif
+
+        default:
+            throw NIOHTTPServerConfigurationError.incompatibleTransportSecurity
```
gjcairo: I wonder if we should throw a test-specific error here instead of reusing this, since it could be mistaken from tests as an error coming from the actual business code.

Tests/NIOHTTPServerTests/Utilities/TestClientConnection.swift
```diff
@@ -0,0 +1,237 @@
...
+    let connectionProtocol: ConnectionProtocol
+
+    /// Asserts the negotiated protocol matches `expectedHTTPVersion`, then returns a request stream.
+    func makeRequestChannel(
+        expectedHTTPVersion: NIOHTTPServer.HTTPVersion,
+        sourceLocation: SourceLocation = #_sourceLocation
+    ) async throws -> NIOAsyncChannel<HTTPResponsePart, HTTPRequestPart> {
+        switch self.connectionProtocol {
+        case .http1(let http1Channel):
+            #expect(
```
gjcairo: These `expect`s in this func should probably be `require`s instead, as anything that we assert after this will fail too anyways.

Tests/NIOHTTPServerTests/Utilities/Helpers.swift
```diff
@@ -94,17 +102,471 @@ extension TLSConfiguration {
...
+        expectedBody: [ByteBuffer],
+        expectedTrailers: HTTPFields? = nil,
+        expectStreamEnd: Bool = true,
+        sourceLocation: SourceLocation = #_sourceLocation
+    ) async throws {
+        var responseIterator = responseStream.makeAsyncIterator()
+
+        for expectedHeadPart in expectedHead {
+            let headResponsePart = try await responseIterator.next()
+            #expect(headResponsePart == .head(expectedHeadPart), sourceLocation: sourceLocation)
```
gjcairo: Similar to my other comment elsewhere, I think these should be `require`s instead of `expect`s

gjcairo (changes requested):

Tests/NIOHTTPServerTests/Utilities/HTTPVersion.swift
aryan-25: Yes, we do, which is why I replaced the `enum HTTPVersion` that existed here before with an extension on the `HTTPVersion` type that exists in the main package.

gjcairo: Oh sorry I missed this was an extension. I thought we had the string value of the protocols (i.e. alps identifiers) there as well, but seems like we don't.

Tests/NIOHTTPServerTests/Utilities/TestClientConnection.swift
aryan-25: Addressed in commit [7e364ed](https://github.com/swift-server/swift-http-server/pull/105/commits/7e364ed57fbfba713645cd22955ec6051279b228).

Tests/NIOHTTPServerTests/Utilities/Helpers.swift
aryan-25: Addressed in commit [7e364ed](https://github.com/swift-server/swift-http-server/pull/105/commits/7e364ed57fbfba713645cd22955ec6051279b228).

Tests/NIOHTTPServerTests/Utilities/TestClientConnection.swift
aryan-25: We need to swallow the `alreadyClosed` error, because in some tests, the connection may have already been closed by the server before we exit from the `body` closure. For example, in `testServerCanContinueDespiteFailedConnection`, the server closes the connection upon seeing the first request. Catching this error at the call site would complicate the tests in my opinion.

Tests/NIOHTTPServerTests/Utilities/TestClientConnection.swift
aryan-25: Addressed in commit [35aec7e](https://github.com/swift-server/swift-http-server/pull/105/commits/35aec7e83eed2b73d9e46f80d1d986ba00110e1b).

Tests/NIOHTTPServerTests/Utilities/NIOClient/NIOClient+HTTP3.swift
aryan-25: Unfortunately we can't remove that because it would mean `self` (which is `QUICConnectionCreator`) would be captured inside the `@Sendable` closure. `QUICConnectionCreator` can't be made `Sendable` as it contains `QUICHandler`.

gjcairo (approved):

PR #106, opened 2026-07-31
Add APIs for reading/writing HTTP Datagrams

aryan-25:
### Motivation

Users may wish to send/receive HTTP Datagrams ([RFC 9297](https://datatracker.ietf.org/doc/html/rfc9297)) using NIOHTTPServer. This PR introduces reader and writer APIs for unreliable datagrams.

Note that the underlying support is yet to be provided by `swift-nio-http3`. The read/write source is not wired up, and we currently throw an error when attempting to read or write to/from the unreliable datagram stream.

### Modifications

- Added an `UnstableHTTPDatagrams` trait, disabled by default. All new APIs are behind that trait (note that the `HTTP3` trait must also be enabled as unreliable datagrams are currently only supported over QUIC).

- Added a `withDatagramReader` method on `NIOHTTPServer.Reader` and a `withDatagramWriter` method on `NIOHTTPServer.ResponseSender.Writer`. Both methods are `consuming` and take a closure that receive the reliable request/response body reader/writer and optionally the unreliable datagram reader/writer as arguments.
  - In both methods, the readers/writers are passed as `consuming sending`, so callers can run the readers/writers separately in concurrent tasks.

- Added `ConnectUDPExample.swift` to show the proposed shape end to end.

### Result

APIs for sending/receiving unreliable HTTP Datagrams added.

### DIFF
--- a/Sources/NIOHTTPServer/Datagrams/NIOHTTPServer+Datagrams.swift
+++ b/Sources/NIOHTTPServer/Datagrams/NIOHTTPServer+Datagrams.swift
@@ -73,21 +73,4 @@ extension NIOHTTPServer.DatagramReader: Sendable {}
 @available(*, unavailable)
 extension NIOHTTPServer.DatagramWriter: Sendable {}
 
-@available(anyAppleOS 26.0, *)
-extension NIOHTTPServer {
-    struct StreamFinish: ~Copyable {
-        let writer: NIOAsyncChannelOutboundWriter<HTTPResponsePart>
-        let state: NIOHTTPServer.ResponseSender.WriterState
-
-        consuming func finish() async throws {
-            // Check if `finish` has not already been fired.
-            let shouldClose = self.state.wrapped.withLock { !$0.finishedWriting }
-            guard shouldClose else { return }
-
-            try await self.writer.write(.end(nil))
-            self.state.wrapped.withLock { $0.finishedWriting = true }
-        }
-    }
-}
-
 #endif  // HTTP3 && UnstableHTTPDatagrams
--- a/Sources/NIOHTTPServer/Datagrams/NIOHTTPServer+Datagrams.swift
+++ b/Sources/NIOHTTPServer/Datagrams/NIOHTTPServer+Datagrams.swift
@@ -1,0 +1,93 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift HTTP Server open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift HTTP Server project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift HTTP Server project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if HTTP3 && UnstableHTTPDatagrams
+
+public import BasicContainers
+public import HTTPAPIs
+import NIOCore
+import NIOHTTPTypes
+import Synchronization
+
+/// Errors from reading/writing on the unreliable datagram.
+@available(anyAppleOS 26.0, *)
+public enum DatagramsError: Error, Sendable {
+    /// The unreliable datagram transport is not yet implemented.
+    case notImplemented
+}
+
+@available(anyAppleOS 26.0, *)
+extension NIOHTTPServer {
+    /// A reader for the unreliable datagram stream.
+    public struct DatagramReader: AsyncReader, ~Copyable {
+        public typealias ReadElement = UInt8
+        public typealias Buffer = UniqueArray<UInt8>
+        public typealias ReadFailure = any Error
+        public typealias FinalElement = Void
+
+        public mutating func read<Return: ~Copyable, Failure: Error>(
+            body: (inout Buffer, consuming FinalElement?) async throws(Failure) -> Return
+        ) async throws(EitherError<ReadFailure, Failure>) -> Return {
+            // TODO: The datagram transport is not yet implemented.
+            throw .first(DatagramsError.notImplemented)
+        }
+    }
+
+    /// A writer for the unreliable datagram stream.
+    public struct DatagramWriter: CallerAsyncWriter, ~Copyable {
+        public typealias WriteElement = UInt8
+        public typealias WriteFailure = any Error
+        public typealias FinalElement = Void
+
+        public mutating func write<Buffer: RangeReplaceableContainer<WriteElement> & ~Copyable>(
+            buffer: inout Buffer
+        ) async throws where Buffer.Element: ~Copyable {
+            // TODO: The datagram transport is not yet implemented.
+            throw DatagramsError.notImplemented
+        }
+
+        public consuming func finish<Buffer: RangeReplaceableContainer<WriteElement> & ~Copyable>(
+            buffer: inout Buffer,
+            finalElement: consuming Void
+        ) async throws where Buffer.Element: ~Copyable {
+            // TODO: The datagram transport is not yet implemented.
+            throw DatagramsError.notImplemented
+        }
+    }
+}
+
+@available(*, unavailable)
+extension NIOHTTPServer.DatagramReader: Sendable {}
+
+@available(*, unavailable)
+extension NIOHTTPServer.DatagramWriter: Sendable {}
+
+@available(anyAppleOS 26.0, *)
+extension NIOHTTPServer {
+    struct StreamFinish: ~Copyable {
+        let writer: NIOAsyncChannelOutboundWriter<HTTPResponsePart>
+        let state: NIOHTTPServer.ResponseSender.WriterState
+
+        consuming func finish() async throws {
+            // Check if `finish` has not already been fired.
+            let shouldClose = self.state.wrapped.withLock { !$0.finishedWriting }
+            guard shouldClose else { return }
+
+            try await self.writer.write(.end(nil))
+            self.state.wrapped.withLock { $0.finishedWriting = true }
+        }
+    }
+}
+
+#endif  // HTTP3 && UnstableHTTPDatagrams
--- a/Sources/NIOHTTPServer/NIOHTTPServer+ConnectionContext.swift
+++ b/Sources/NIOHTTPServer/NIOHTTPServer+ConnectionContext.swift
@@ -66,25 +66,6 @@ extension NIOHTTPServer {
             self.peerCertificateChainFuture = peerCertificateChainFuture
         }
 
-        #if HTTP3 && UnstableHTTPDatagrams
-        /// Whether this connection supports unreliable datagrams.
-        public var supportsUnreliableDatagrams: Bool {
-            switch self.httpVersion {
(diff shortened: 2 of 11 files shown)


Sources/NIOHTTPServer/NIOHTTPServerReader.swift
```diff
@@ -69,14 +69,31 @@ extension NIOHTTPServer {
...
         init(readerState: ReaderState) {
             self.state = readerState
             self.iterator = readerState.takeIterator()
             self.buffer = UniqueArray<UInt8>()
         }
 
+        #if HTTP3 && UnstableHTTPDatagrams
+        /// The unreliable datagram reader, present when the underlying transport is capable of reading/writing
+        /// unreliable datagrams.
+        private var datagramReader: Disconnected<NIOHTTPServer.DatagramReader>?
```
gjcairo: Do we need `Disconnected`?

Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
```diff
@@ -107,6 +142,41 @@ extension NIOHTTPServer.ResponseSender {
...
+        ///
+        /// - Parameter body: A closure that receives the reliable stream writer and the datagram writer. The datagram
+        ///   writer is `nil` when the underlying transport does not support unreliable datagrams.
+        ///
+        /// - Note: Both writers are passed as `consuming sending`. This means that `body` can use both writers in
+        ///   separate tasks.
+        public consuming func withDatagramWriter(
+            _ body: (consuming sending Self, consuming sending NIOHTTPServer.DatagramWriter?) async throws -> Void
+        ) async throws {
+            let streamFinish = NIOHTTPServer.StreamFinish(writer: self.writer, state: self.writerState)
```
gjcairo: I think this type is a bit unnecessary and adds some indirection which makes things harder to follow, IMO. I'd just finish both streams inline in this func, like we do elsewhere.

Tests/NIOHTTPServerTests/NIOHTTPServerReaderTests.swift
```diff
@@ -204,4 +204,67 @@ struct NIOHTTPServerReaderTests {
...
+    @available(anyAppleOS 26.0, *)
+    func withDatagramReaderVendsRequestAndDatagramReader() async throws {
+        let (stream, source) = NIOAsyncChannelInboundStream<HTTPRequestPart>.makeTestingStream()
+        source.yield(.body(ByteBuffer(bytes: [1, 2, 3])))
+        source.yield(.end(nil))
+        source.finish()
+
+        let reader = NIOHTTPServer.Reader(
+            readerState: .init(iterator: stream.makeAsyncIterator()),
+            datagramReader: .init(value: .init())
```
gjcairo: All these `.init`s are a bit confusing to understand - can we spell out the names?

gjcairo (changes requested):

Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
aryan-25: We unfortunately need this indirection because we pass the stream writer (and the datagram writer) as `consuming sending` to the `body` closure, which means we can't call `finish` on either after the `body` closure.

Sources/NIOHTTPServer/NIOHTTPServerReader.swift
aryan-25: Yes, without this, the datagramReader can't be vended out as `sending` in `withDatagramReader`.

Tests/NIOHTTPServerTests/NIOHTTPServerReaderTests.swift
aryan-25: Addressed in commit [5986cd4](https://github.com/swift-server/swift-http-server/pull/106/commits/5986cd438fb8753f0d651e77c3b5be1169336e12).

Sources/NIOHTTPServer/NIOHTTPServerReader.swift
gjcairo: I see. 

I am quite unconvinced about storing this reader inside the stream reader to be honest - I think it would make more sense to have this somewhere else, like the request context. But I'm also aware this will likely be temporary until we can come up with a more general solution to CONNECT.

Sources/NIOHTTPServer/NIOHTTPServerReader.swift
```diff
@@ -121,3 +138,29 @@ extension NIOHTTPServer {
...
+@available(anyAppleOS 26.0, *)
+extension NIOHTTPServer.Reader {
+    /// Vends this request body reader and the unreliable datagram reader to the provided `body` closure.
+    ///
+    /// - Parameter body: A closure that receives the request body reader and the datagram reader. The datagram reader
+    ///   is `nil` when the underlying transport does not support unreliable datagrams.
+    ///
+    /// - Note: Both readers are passed as `consuming sending`. This means that `body` can use both readers in separate
+    ///   tasks.
+    public consuming func withDatagramReader(
```
gjcairo: Related to my last comment, some things about this function I'm not sure about. The fact we need to create a new `Reader` isn't great. Users would also be able to call `withDatagramReader` again on the `streamReader` that's returned (since it's the same type), which is a bit weird. Again, perhaps it's fine since this will be temporary, but I thought I'd point them out.

Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
gjcairo: This is actually making me doubt we need to have a with-style method (also on the reader). I don't think we should finish the writers implicitly - we don't do this for the normal stream writer: we just expose `write` and `finish`. Why not return a tuple with both writers (and readers) instead of having a closure?

aryan-25: Regardless of whether we should automatically finish or not

> Why not return a tuple with both writers (and readers) instead of having a closure?

This won't work because tuples with non-Copyable elements aren't supported yet.

I think the closure shape is the only shape that allows both the readers/writers to be run in concurrent tasks.

Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
```diff
@@ -144,37 +166,14 @@ extension NIOHTTPServer.ResponseSender {
         }
 
         #if HTTP3 && UnstableHTTPDatagrams
-        /// Vends this reliable stream writer and the unreliable datagram writer to the provided `body` closure, and
-        /// finishes the reliable stream afterwards, only if not already finished by the caller in the `body` closure.
+        /// Returns the unreliable datagram writer for this stream, if there is one.
         ///
-        /// - Parameter body: A closure that receives the reliable stream writer and the datagram writer. The datagram
-        ///   writer is `nil` when the underlying transport does not support unreliable datagrams.
+        /// - Returns: The unreliable datagram writer for this stream if one is available.
```
aryan-25: This is conveying the same thing as the line above. We can remove this line.

Tests/NIOHTTPServerTests/NIOHTTPServerReaderTests.swift
```diff
@@ -206,63 +206,61 @@ struct NIOHTTPServerReaderTests {
     }
 
     #if HTTP3 && UnstableHTTPDatagrams
-    @Test("withDatagramReader vends no datagram reader when not available")
+    @Test("takeDatagramReader vends no datagram reader when not available")
     @available(anyAppleOS 26.0, *)
     func withDatagramReaderVendsNilWhenNotAvailable() async throws {
```
aryan-25: We should also rename the function names to be consistent with the new `takeDatagram{Reader|Writer}` naming. The same applies for the tests below and also in `NIOHTTPServerWriterTests`.

Sources/NIOHTTPServer/Datagrams/NIOHTTPServer+Datagrams.swift
```diff
@@ -0,0 +1,93 @@
...
+
+@available(*, unavailable)
+extension NIOHTTPServer.DatagramReader: Sendable {}
+
+@available(*, unavailable)
+extension NIOHTTPServer.DatagramWriter: Sendable {}
+
+@available(anyAppleOS 26.0, *)
+extension NIOHTTPServer {
+    struct StreamFinish: ~Copyable {
```
aryan-25: We can now also remove this.

Sources/NIOHTTPServer/NIOHTTPServerReader.swift
```diff
@@ -121,3 +138,17 @@ extension NIOHTTPServer {
 
 @available(*, unavailable)
 extension NIOHTTPServer.Reader: Sendable {}
+
+#if HTTP3 && UnstableHTTPDatagrams
+@available(anyAppleOS 26.0, *)
+extension NIOHTTPServer.Reader {
+    /// Returns the unreliable datagram reader for this stream, if there is one.
+    ///
+    /// - Returns: The unreliable datagram reader for this stream if one is available.
```
aryan-25: This is conveying the same thing as the line above. We can remove this line.

Sources/NIOHTTPServer/NIOHTTPServer+ConnectionContext.swift
```diff
@@ -66,6 +66,25 @@ extension NIOHTTPServer {
             self.peerCertificateChainFuture = peerCertificateChainFuture
         }
 
+        #if HTTP3 && UnstableHTTPDatagrams
+        /// Whether this connection supports unreliable datagrams.
+        public var supportsUnreliableDatagrams: Bool {
```
aryan-25: We probably don't need this (and the mirrored accessor on `RequestContext`) anymore.

aryan-25 (commented):
The revised shape looks good. Just some small comments. Thanks for driving this forward @gjcairo!

ehaydenr (approved):

aryan-25:
LGTM, thank you!

Sources/NIOHTTPServer/Datagrams/NIOHTTPServer+Datagrams.swift
```diff
@@ -0,0 +1,76 @@
...
+public enum DatagramsError: Error, Sendable {
+    /// The unreliable datagram transport is not yet implemented.
+    case notImplemented
+}
+
+@available(anyAppleOS 26.0, *)
+extension NIOHTTPServer {
+    /// A reader for the unreliable datagram stream.
+    public struct DatagramReader: AsyncReader, ~Copyable {
+        public typealias ReadElement = UInt8
```
guoye-zhang, line 34: Should the element be a datagram instead of a byte?

gjcairo: Could be, although I wonder if we want our own type here or if (at least for now) we'd be okay reusing NIOH3's Datagram type. I'll keep this as is right now and we can decide when we integrate with NIOH3.

gjcairo (approved):

PR #107, opened 2026-08-03
Validate configuration at initialization

aryan-25:
### Motivation

The required TLS contexts/configurations required to create the server channels are only constructed when `NIOHTTPServer.serve` is invoked. The problem with this is that these constructions can throw an error if the provided configuration was invalid.

We should not allow an invalid `NIOHTTPServerConfiguration` instance to be created in the first place: validation should occur during the initialization of `NIOHTTPServerConfiguration`.

### Modifications

- `NIOHTTPServerConfiguration`'s initializer now validates that `supportedHTTPVersions` and `transportSecurity` are compatible with each other, and stores the necessary TLS objects.
  - `NIOHTTPServer.serve .serve` can therefore use these properties directly and not invoke any `throw`ing method relating to configuration.
- Some small general refactors to improve readability.

### Result

Validation of `NIOHTTPServerConfiguration` now happens at initialization rather than during `NIOHTTPServer.serve`.


Tests/NIOHTTPServerTests/NIOHTTPServerConfigurationTests.swift
```diff
@@ -0,0 +1,388 @@
...
+        }
+        #endif  // HTTP3
+    }
+
+    @Suite
+    struct TransportSecurity {
+        @available(anyAppleOS 26.0, *)
+        @Test(
+            "All X.509 credential sources produce a valid configuration",
+            arguments: [TestX509CredentialSource]([.inMemory, .reloading, .pemFile, .derFile, .pemBytes, .derBytes])
```
gjcairo: Nit, but for consistency with other tests/files/packages:
```suggestion
            arguments: [TestX509CredentialSource.inMemory, .reloading, .pemFile, .derFile, .pemBytes, .derBytes]
```
This has to change in a few places.

Tests/NIOHTTPServerTests/NIOHTTPServerConfigurationTests.swift
```diff
@@ -0,0 +1,388 @@
...
+            "A non-existent X.509 certificate file path is rejected",
+            arguments: [NIOHTTPServerConfiguration.TransportSecurity.X509Credentials]([
+                .pemFile(certificateChainPath: "/does/not/exist.pem", privateKeyPath: "/does/not/exist.key"),
+                .derFile(certificatePath: "/does/not/exist.der", privateKeyPath: "/does/not/exist.der"),
+            ])
+        )
+        func nonExistentX509FilePathRejected(
+            credentials: NIOHTTPServerConfiguration.TransportSecurity.X509Credentials
+        ) throws {
+            #expect(throws: Error.self) {
```
gjcairo, line 144: We can't assert a more specific error type or message?

Tests/NIOHTTPServerTests/NIOHTTPServerConfigurationTests.swift
```diff
@@ -0,0 +1,388 @@
...
+                    )
+                )
+            }
+        }
+        #endif  // HTTP3
+    }
+}
+
+@available(anyAppleOS 26.0, *)
+enum TestX509CredentialSource: Sendable {
```
gjcairo: We don't have a type like this in the actual configuration types? Feels a bit weird to have this parallel type in the tests instead of an extension on that type similarly to what we're doing below with `NIOHTTPServerConfiguration.TransportSecurity.RawPublicKeyCredentials`.

Tests/NIOHTTPServerTests/NIOHTTPServerConfigurationTests.swift
```diff
@@ -0,0 +1,388 @@
...
+        try Data(try publicKey.serializeAsPEM().derBytes).write(to: publicKeyPath)
+        try Data(try privateKey.serializeAsPEM().derBytes).write(to: privateKeyPath)
+
+        return .derFile(publicKeyPath: publicKeyPath.path, privateKeyPath: privateKeyPath.path)
+    }
+}
+#endif  // HTTP3
+
+@available(anyAppleOS 26.0, *)
+enum MTLSTrustSource: Sendable {
```
gjcairo: Same question here

Tests/NIOHTTPServerTests/NIOHTTPServerConfigurationTests.swift
```diff
@@ -0,0 +1,388 @@
...
+}
+
+@available(anyAppleOS 26.0, *)
+extension NIOHTTPServerConfiguration.HTTPVersion {
+    static let http2 = Self.http2(config: .defaults)
+
+    #if HTTP3
+    static let http3 = Self.http3(config: .defaults)
+    #endif
+}
```
gjcairo: Perhaps we should default `config` to `.defaults` in the `init` instead of having this

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
```diff
@@ -285,21 +285,23 @@ public struct NIOHTTPServerConfiguration: Sendable {
     }
 
     /// Network binding configuration specifying all addresses where the server should listen.
-    public var bindTargets: [BindTarget]
+    public let bindTargets: [BindTarget]
```
gjcairo: We generally keep properties in config structs as vars so they can be modified. This is common across all of our packages AFAIK. Did you make this change for a specific reason?

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration+Validation.swift
```diff
@@ -0,0 +1,110 @@
...
+import NIOSSL
+
+#if HTTP3
+import NIOQUIC
+#endif
+
+@available(anyAppleOS 26.0, *)
+extension NIOHTTPServerConfiguration {
+    /// The context required to serve a secure upgrade channel.
+    struct SecureUpgradeContext {
```
gjcairo: Can we rename to `ValidatedSecureUpgradeConfiguration` perhaps? Same for HTTP3 below.

gjcairo (changes requested):

Tests/NIOHTTPServerTests/NIOHTTPServerConfigurationTests.swift
aryan-25: Different errors are returned in both cases by `swift-nio-ssl`. The PEM case throws a `NIOSSLError.failedToLoadCertificate` error and the DER case throws a `NIOCore.IOError`.

I felt that explicitly catching these errors clutters the test case without adding much benefit.

Tests/NIOHTTPServerTests/NIOHTTPServerConfigurationTests.swift
aryan-25: The actual configuration types don't allow us to express the cases (e.g. `inMemory`, `reloading`, etc.) without providing associated values.

These parallel types allow us to parameterise over all cases quite easily. There is only one supported format for RPK credentials (DER files), which is why I didn't bother for that case.

Tests/NIOHTTPServerTests/NIOHTTPServerConfigurationTests.swift
aryan-25: Answered [here](https://github.com/swift-server/swift-http-server/pull/107#discussion_r3721529985).

Tests/NIOHTTPServerTests/NIOHTTPServerConfigurationTests.swift
aryan-25: Addressed in commit [6605bc2](https://github.com/swift-server/swift-http-server/pull/107/commits/6605bc2ca30f0206746c799b9f8a0492de6a7918).

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
aryan-25: Yes. If any of these three properties (`bindTargets`, `transportSecurity`, `supportedHTTPVersions`) are modified after initialization, then the two new contexts we store may become invalid.

Tests/NIOHTTPServerTests/NIOHTTPServerConfigurationTests.swift
aryan-25: Addressed in commits [8da01f8](https://github.com/swift-server/swift-http-server/pull/107/commits/8da01f8193938f74f1ec52f7165d465bedecad6e) and [bd7757d](https://github.com/swift-server/swift-http-server/pull/107/commits/bd7757dd2b12ef4bd16cb0c370699a6259ffdb5e).

Tests/NIOHTTPServerTests/NIOHTTPServerConfigurationTests.swift
gjcairo: I only worry that expecting _any_ error may mean that even though this behaviour changes in the future (e.g. a new error we should be handling different starts being thrown), the test will still pass.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
gjcairo: We can add a `willSet` to these properties and do the validation/update of the contexts there.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
```diff
@@ -523,15 +523,15 @@ extension NIOHTTPServerConfiguration {
         /// The HTTP/2 protocol version.
         ///
         /// - Parameter config: The configuration to use for HTTP/2.
-        public static func http2(config: HTTP2) -> Self {
+        public static func http2(config: HTTP2 = .defaults) -> Self {
```
gjcairo, line 571: Nit, but I think additionally having a `static var` for h2 and h3 that defaults the config to `.defaults` may make the API slightly nicer since we can avoid the empty brackets (`.http2` instead of `.http2()`)

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
gjcairo: We discussed offline and because we can't have throwing setters we will probably just precondition failure.
Thinking more about it though, because this wouldn't fail on release builds, I think it's better to avoid having the `ValidatedX` types and instead just do validations in the init/setters, without encapsulating in separate types.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
aryan-25: Addressed in commit [b32144d](https://github.com/swift-server/swift-http-server/pull/107/commits/b32144dbe66b9edeb2eb561cc738d60a7adfcdc3).

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
aryan-25: I've now updated the PR with this approach.

FranzBusch:
This is interesting. A potential side observation: This seems to boil down to differences in support transport security for H1/2/3. It would be great if this is a compile time guarantee rather than a runtime enforced. Have we considered making the transport security structs HTTP version specific?

aryan-25:
> This seems to boil down to differences in support transport security for H1/2/3

Yes. Although it is worth nothing that most of the differences today just arise from a mismatch in the credential sources supported by `swift-nio-ssl` and `swift-nio-quic`. From `swift-nio-quic`, we need support for (1) byte-array, DER-file, and in-memory `swift-certificates` sources, (2) certificate reloading, and (3) mTLS (with support for a custom verification callback). From `swift-nio-ssl`, we need support for Raw Public Key sources.

Once there is parity, the only difference that remains is `transportSecurity == .plaintext` being possible over HTTP/1.1 (and HTTP/2 once we add support for h2c), but not over HTTP/3.

It feels odd to have a `transportSecurity` type for HTTP/3 and a very similar one for HTTP/1.1 / HTTP/2 that additionally just allows `plaintext` to be expressed. Having a `transportSecurity` per supported HTTP version may not also be the right shape because when `supportedHTTPVersions == [.http1_1, .http2]`, peers may be intending to use either HTTP/1.1 or HTTP/2, but we only find out through the TLS handshake (ALPN), which means we can only support a single set of TLS credentials.

Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift
```diff
@@ -285,13 +289,47 @@ public struct NIOHTTPServerConfiguration: Sendable {
...
+    ///
+    /// - Precondition: Must be compatible with ``supportedHTTPVersions``:
+    ///   - `transportSecurity == .mTLS` is not supported when `supportedHTTPVersions` contains `.http3`.
+    ///   - Raw Public Key credentials are only supported when `supportedHTTPVersions == [.http3]`.
+    ///   - When `supportedHTTPVersions` contains `.http2` and `.http3`, TLS credentials must be provided as PEM files
+    ///     on disk. Other credential sources are not supported.
+    ///   - `transportSecurity` can only be set to `.plaintext` when `supportedHTTPVersions == [.http1_1]`.
+    public var transportSecurity: TransportSecurity {
+        didSet {
+            try! self.validateTransportConfiguration()
```
gjcairo: I don't think we should `try!` here. I think it's okay to precondition failure because we won't crash at runtime on release builds, but with this we could. We should `throw` in throwing contexts and precondition failure where we can't throw.

aryan-25: Note that `preconditionFailure` will also crash in release builds (assuming standard release build configs). It will only _not_ crash when the package is built with `-Ounchecked`, which is a non-standard release build config: https://developer.apple.com/documentation/swift/preconditionfailure(_:file:line:)#discussion.

But I've removed the `try!` and replaced it with `preconditionFailure` for consistency in commit 5c29ae1.

gjcairo (approved):

FranzBusch:
> Yes. Although it is worth nothing that most of the differences today just arise from a mismatch in the credential sources supported by `swift-nio-ssl` and `swift-nio-quic`. From `swift-nio-quic`, we need support for (1) byte-array, DER-file, and in-memory `swift-certificates` sources, (2) certificate reloading, and (3) mTLS (with support for a custom verification callback). From `swift-nio-ssl`, we need support for Raw Public Key sources.
> 
> Once there is parity, the only difference that remains is `transportSecurity == .plaintext` being possible over HTTP/1.1 (and HTTP/2 once we add support for h2c), but not over HTTP/3.
> 
> It feels odd to have a `transportSecurity` type for HTTP/3 and a very similar one for HTTP/1.1 / HTTP/2 that additionally just allows `plaintext` to be expressed. Having a `transportSecurity` per supported HTTP version may not also be the right shape because when `supportedHTTPVersions == [.http1_1, .http2]`, peers may be intending to use either HTTP/1.1 or HTTP/2, but we only find out through the TLS handshake (ALPN), which means we can only support a single set of TLS credentials.

I understand where the current differences come from. Just to explain this a bit more. What if a user wants to bootstrap a server that serves H/1 over plaintext and H2 over TLS? This is currently not configurable right? This might be fine just trying to provide another argument why separate H1/2/3/ properties might be beneficial. I agree though that they should use the same type.

aryan-25:
> What if a user wants to bootstrap a server that serves H/1 over plaintext and H2 over TLS? This is currently not configurable right?

Correct, this is not currently configurable.

> This might be fine just trying to provide another argument why separate H1/2/3/ properties might be beneficial. I agree though that they should use the same type.

I think we should think more about how we can best model this. There are certain combinations that we may not want to support, e.g. HTTP/1.1 over TLS and HTTP/2 over plaintext. Should we move this discussion to a GH issue?

FranzBusch:
> I think we should think more about how we can best model this. There are certain combinations that we may not want to support, e.g. HTTP/1.1 over TLS and HTTP/2 over plaintext. Should we move this discussion to a GH issue?

Happy to move to a GH issue. Though I think both of the configurations you mentioned are entirely valid and we must support them. It is common to serve an HTTP/2 server over plaintext in data center environments.

aryan-25:
> Though I think both of the configurations you mentioned are entirely valid and we must support them.

Sorry, I meant to convey both of these specified as one single unit, i.e. at the start of every connection, the server would need to determine if the peer is trying to perform HTTP/1.1 over TLS _OR_ h2c, and then add the correct channel handlers accordingly.

I have created the issue here: #111.

PR #109, opened 2026-08-11
Disable Swift Configuration Traits

0xTim:
Swift Configuration's default traits pull in full Foundation because they enable the JSON configuration provider, which uses JSONSerialization. That undoes all the hard work in the library to only use FoundationEssentials as full foundation will be linked anyway. Since the server doesn't need any of the trait-hidden code, let's not include them


gjcairo:
Thanks @0xTim !

gjcairo (approved):

gjcairo:
Looks like CI is failing because of an SPM bug: https://github.com/swiftlang/swift-package-manager/issues/10397

aryan-25 (approved):

FranzBusch (approved):

PR #110, opened 2026-08-12
Raise QUIC and HTTP/3 dependencies

aryan-25:
### Motivation

`swift-network-evolution`'s 0.1.2 release introduced API breaking changes. This resulted in `swift-nio-quic` failing to build, which in turn also breaks `swift-http-server`'s HTTP/3 support. This was however fixed in `swift-nio-quic`'s [0.2.1 release](https://github.com/apple/swift-nio-quic/releases/tag/0.2.1).

### Modifications

- Raised the `swift-nio-quic` dependency to v0.2.1.
- Raised the `swift-nio-http3` dependency to v0.2.0.
- Adopted an API breaking change in the test HTTP/3 client.

### Result

This package can build successfully when the HTTP/3 trait is enabled.


Tests/NIOHTTPServerTests/Utilities/NIOClient/NIOClient+HTTP3.swift
```diff
@@ -44,7 +44,9 @@ struct QUICConnectionCreator: HTTP3ConnectionCreator {
                 }
             },
             inboundStreamInitializer: self.inboundStreamInitializer
-        )
+        ).map { connectionChannel, _ in
+            connectionChannel
+        }
```
gjcairo, line 49: What is the second element of the tuple? The multiplexer?

aryan-25: The second element is the `QUICStreamCreator`. We don't need it here.

gjcairo (approved):

FranzBusch (approved):

PR #115, opened 2026-08-18
Reset streams when throwing from request handler

gjcairo:
This change translates errors thrown from the request handler into stream resets.

Two new protocols have been added: `NIOHTTPServerHTTP2StreamResetError` and `NIOHTTPServerHTTP3StreamResetError`, which allow to specify how custom errors map to HTTP/2 stream reset error codes and and HTTP/3 stream reset and stop sending error codes, respectively.

The behaviour of throwing an error changes depending on which protocol the request is being served over:
- HTTP/1: there are no streams, so the connection is closed. If the head has not yet been sent, it will be amended to include `Connection: close`. 
- HTTP/2: `RST_STREAM` is sent over the wire, closing the stream bidirectionally. If the error thrown conforms to `NIOHTTPServerHTTP2StreamResetError`, the mapped error code will be used; if not, the code will be defaulted to `INTERNAL_ERROR` (0x02).
- HTTP/3: `STREAM_RESET` and `STOP_SENDING` are both sent to the client, closing the stream bidirectionally. If the error thrown conforms to `NIOHTTPServerHTTP3StreamResetError`, the mapped error code will be used; if not, the code will be defaulted to `H3_INTERNAL_ERROR` (0x0102).


Sources/NIOHTTPServer/NIOHTTPServer+ConnectionContext.swift
```diff
@@ -19,7 +19,7 @@ public import X509
 @available(anyAppleOS 26.0, *)
 extension NIOHTTPServer {
     /// The application-level HTTP version negotiated for a connection.
-    public enum HTTPVersion: String, Sendable, Hashable {
+    public enum HTTPVersion: String, Sendable, Hashable, CaseIterable {
```
FranzBusch, line 23: We need to mark this as `@nonexhaustive` otherwise we won't be able to support a hypothethical `HTTP/4`.

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -354,19 +376,32 @@ public struct NIOHTTPServer: HTTPServer {
...
             )
         } catch {
-            logger.error("Error thrown while handling request: \(error)")
-            if !readerState.wrapped.withLock({ $0.finishedReading }) {
-                logger.error("Did not finish reading but error thrown.")
-            }
+            // A throwing handler signals that the exchange failed. The error is deliberately not propagated to any
+            // caller: it exists to drive the wire, aborting the exchange with protocol error codes the error can choose
+            // by conforming to `NIOHTTPServerHTTP2StreamResetError` / `NIOHTTPServerHTTP3StreamResetError`.
+            self.logger.error(
```
FranzBusch: This must not log at error level. Libraries should only log at info level and below. I think this is still somewhat the normal case so I would recommend logging at trace or debug level.

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -354,19 +376,32 @@ public struct NIOHTTPServer: HTTPServer {
...
+                "Error thrown while handling request: aborting.",
+                metadata: [
+                    "error": "\(error)",
```
FranzBusch: swift-log has new APIs that you can pass an error to. Can we use that instead?

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -354,19 +376,32 @@ public struct NIOHTTPServer: HTTPServer {
...
+                    "protocol": "\(requestContext.connectionContext.httpVersion)",
```
FranzBusch: I would recommend that we move all the keys into a internal struct called `LoggingKeys` so we don't have random string literals floating around

Sources/NIOHTTPServer/NIOHTTPServerHTTP2StreamResetError.swift
```diff
@@ -0,0 +1,44 @@
...
+/// }
+///
+/// try await server.serve { request, context, reader, responseSender in
+///     guard let tunnel = try? await openTunnel(to: request.authority) else {
+///         throw TunnelFailure()
+///     }
+///     // ...
+/// }
+/// ```
+public protocol NIOHTTPServerHTTP2StreamResetError: Error {
```
FranzBusch: I would recommend a different name here following what GRPC does https://github.com/grpc/grpc-swift-2/blob/76daaeaf345213fbd1b0ae78121c06036a0ddbfe/Sources/GRPCCore/RPCError.swift#L309
```suggestion
public protocol NIOHTTPServerHTTP2StreamResetErrorConvertible: Error {
```

Sources/NIOHTTPServer/NIOHTTPServerHTTP3StreamResetError.swift
```diff
@@ -0,0 +1,55 @@
...
+/// }
+///
+/// try await server.serve { request, context, reader, responseSender in
+///     guard let tunnel = try? await openTunnel(to: request.authority) else {
+///         throw TunnelFailure()
+///     }
+///     // ...
+/// }
+/// ```
+public protocol NIOHTTPServerHTTP3StreamResetError: Error {
```
FranzBusch: Same here
```suggestion
public protocol NIOHTTPServerHTTP3StreamResetErrorConvertible: Error {
```

Sources/NIOHTTPServer/NIOHTTPServerHTTP2StreamResetError.swift
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift HTTP Server project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+public import NIOHTTP2
```
FranzBusch: I'm not happy about this public import. I would prefer if we can continue to avoid exporting NIO into our public API where possible. We should find a better place for this error to live. Maybe even this package

Sources/NIOHTTPServer/NIOHTTPServerHTTP3StreamResetError.swift
```diff
@@ -0,0 +1,55 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift HTTP Server project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if HTTP3
+public import HTTP3
```
FranzBusch: Same here

Sources/NIOHTTPServer/NIOHTTPServer.swift
gjcairo: I've changed this and also updated other logs where we were using `error`

Sources/NIOHTTPServer/NIOHTTPServer.swift
gjcairo: I updated other instances to use the new API too

Sources/NIOHTTPServer/NIOHTTPServerHTTP2StreamResetError.swift
gjcairo: Fair. I changed the type here to `UInt32` to remove the NIO dependency and to leave it up to implementers to decide where they get the errors codes from (I've added an example too showing you can get the errors from the NIO H2 error codes' raw value). Let me know what you think. Like this, we could upstream these protocols to the API proposal package eventually. 

(I know we generally expose `Int` in APIs, but only valid values for H2 steam error codes are positive integers up to 32 bits so it seems better to enforce it like this).

Sources/NIOHTTPServer/NIOHTTPServerHTTP3StreamResetError.swift
gjcairo: I did the same here, but using UInt64, for the same reasons as I explained above.

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
```diff
@@ -435,11 +446,9 @@ extension NIOHTTPServer {
                             return .certificateVerified(.init(.init(nioSSLCerts)))
 
                         case .failed(let error):
-                            self.logger.error(
+                            self.logger.debug(
                                 "Custom certificate verification failed",
-                                metadata: [
-                                    "failure-reason": .string(error.reason)
-                                ]
+                                metadata: [LoggingKeys.failureReason: .string(error.reason)]
```
FranzBusch: Why don't we pass the error to the log here. Then we can remove that metadata completely

FranzBusch (approved):
LGTM. One minor nit

Sources/NIOHTTPServer/NIOHTTPServer+AbortRequest.swift
```diff
@@ -0,0 +1,84 @@
...
+    /// `Connection: close` handling.
+    /// - HTTP/2 and HTTP/3 reset the request's own stream, with the error codes `error` describes.
+    /// - HTTP/3 also asks the client to STOP_SENDING.
+    static func abortRequest(requestContext: RequestContext, error: any Error) {
+        let channel = requestContext.channel
+
+        switch requestContext.connectionContext.httpVersion {
+        case .plaintextHTTP1_1, .http1_1:
+            var response = HTTPResponse(status: .internalServerError)
+            response.headerFields[.contentLength] = "0"
```
aryan-25, line 41: We don't need to do this, right?

Tests/NIOHTTPServerTests/NIOHTTPServerStreamResetTests.swift
```diff
@@ -0,0 +1,493 @@
...
+            clientConfiguration: clientConfiguration,
+            server: server,
+            serverHandler: HTTPServerClosureRequestHandler { _, _, _, _ in
+                throw StreamResetTestError.connectFailed
+            }
+        ) { _, inbound, outbound in
+            try await outbound.write(.testHead(method: .get, for: http1Variant))
+            try await outbound.write(.end(nil))
+
+            var responseIterator = inbound.makeAsyncIterator()
```
aryan-25: We can use `TestHelpers.validateResponse` here with `expectStreamEnd` set to `true`.

Sources/NIOHTTPServer/HTTPServerHTTP2StreamResetErrorConvertible.swift
```diff
@@ -0,0 +1,47 @@
...
+/// }
+/// ```
+public protocol HTTPServerHTTP2StreamResetErrorConvertible: Error {
+    /// The `RST_STREAM` error code to send, as its numeric value on the wire.
+    ///
+    /// The codes and their values are defined by RFC 9113 § 7 — for example `0x08` for `CANCEL`, `0x0a` for
+    /// `CONNECT_ERROR`, or `0x02` for `INTERNAL_ERROR`.
+    ///
+    /// This code is used only when the request is served over HTTP/2.
+    var http2StreamResetCode: UInt32 { get }
```
ehaydenr, line 46: Did you consider using https://github.com/apple/swift-nio-http2/blob/281170341aeff301a62dab0ba13f9829ff148989/Sources/NIOHTTP2/HTTP2ErrorCode.swift#L21 instead of a raw UInt32?

Tests/NIOHTTPServerTests/NIOHTTPServerStreamResetTests.swift
```diff
@@ -0,0 +1,493 @@
...
+    @available(anyAppleOS 26.0, *)
+    func testHTTP3ThrowingResetsStream() async throws {
+        let (server, clientConfiguration) = try TestHelpers.makeServerAndClientConfiguration(
+            for: .http3,
+            clientLogger: self.clientLogger,
+            serverLogger: self.serverLogger
+        )
+
+        // The stream is reset rather than completed, so draining the response fails instead of ending cleanly.
+        await #expect(throws: (any Error).self) {
```
aryan-25: You can do this here:

```suggestion
        let error = try await #require(throws: HTTP3Error.self) {
```

and then assert the `connectError`:
```
#expect(error.code == .remoteStreamError)
#expect(error.h3ErrorCode == .connectError)
```

Tests/NIOHTTPServerTests/NIOHTTPServerStreamResetTests.swift
```diff
@@ -0,0 +1,493 @@
...
+    @available(anyAppleOS 26.0, *)
+    func testHTTP3ThrowingAfterResponseHeadResetsStream() async throws {
+        let (server, clientConfiguration) = try TestHelpers.makeServerAndClientConfiguration(
+            for: .http3,
+            clientLogger: self.clientLogger,
+            serverLogger: self.serverLogger
+        )
+
+        // The response is abandoned after its head, so draining it fails instead of ending cleanly.
+        await #expect(throws: (any Error).self) {
```
aryan-25: Same applies here.

Tests/NIOHTTPServerTests/NIOHTTPServerStreamResetTests.swift
```diff
@@ -0,0 +1,493 @@
...
+            serverHandler: HTTPServerClosureRequestHandler { _, _, _, sender in
+                // Conclude the response fully, then fail.
+                try await sender.sendAndFinish(.init(status: .ok, headerFields: [.contentLength: "0"]))
+                throw StreamResetTestError.connectFailed
+            }
+        ) { _, inbound, outbound in
+            try await outbound.write(.testHead(method: .get, for: httpVersion))
+            try await outbound.write(.end(nil))
+
+            var responseIterator = inbound.makeAsyncIterator()
```
aryan-25: This block can be replaced with `TestHelpers.validateResponse` with `expectStreamEnd` set to `true`.

Sources/NIOHTTPServer/HTTPServerHTTP2StreamResetErrorConvertible.swift
gjcairo: See https://github.com/swift-server/swift-http-server/pull/115#discussion_r3804231589

Sources/NIOHTTPServer/NIOHTTPServer+AbortRequest.swift
gjcairo: Which bit? The content length? If there is no body in the response, servers should set content length to 0.

ehaydenr (approved):

aryan-25 (approved):

PR #117, opened 2026-08-21
Wire up unreliable datagram transport

aryan-25:
### Motivation

In #106, we introduced high-level APIs for reading/writing unreliable HTTP/3 datagrams. Since then, `swift-nio-quic` and `swift-nio-http3` added support for reading/writing unreliable datagrams. This PR wires up the high-level APIs to the underlying QUIC and HTTP/3 datagram APIs. 

### Modifications

- Added `HTTP3ConnectionManager`, which is a channel handler that is placed after `HTTP3ConnectionHandler`. This handler simply just reads the stream ID from the `HTTP3Datagram` and routes the datagram to the request stream (if registered).
  - This handler also serves as the handler to perform other connection management logic (such as closing the connection upon a connection error), hence named that way.

- Added `HTTP3UnreliableDatagramStream`, which wraps over an `AsyncStream` and is the source/sink for the datagram reader and writer that are vended in the request stream. This type also buffers a bounded (and configurable) number of datagrams.

- Added datagram configuration types.

- Updated the `takeDatagramReader` and `takeDatagramWriter` methods to be `async`. This is because:
  - [RFC 9297 § 2.1.1](https://datatracker.ietf.org/doc/html/rfc9297#section-2.1.1-3) states that datagrams must not be sent until the endpoint has received the SETTINGS_H3_DATAGRAM setting from the peer.
  - But at the time of initializing the request stream (i.e. creating the datagram reader/writer to vend to the request handler), it is not guaranteed that the SETTINGS frame (on the control frame) has arrived (QUIC streams are unordered relative to one another).
  - In the stream initializer, we could wait until the SETTINGS frame has arrived and only then invoke the request handler. But this goes against what [RFC 9114 § 7.2.4.2](https://datatracker.ietf.org/doc/html/rfc9114#section-7.2.4-2) recommends (streams should not be blocked on SETTINGS arriving).
  - So in this PR, I've made `takeDatagramReader` and `takeDatagramWriter` suspend until the peer's SETTINGS has arrived. This means we only impose blocking on streams where users intend to read/write datagrams on.

  - `takeDatagramReader`/`takeDatagramWriter` now return `nil` in the following scenarios:
    1. If the client does not agree to receiving datagrams, or;
    2. If datagrams were explicitly disabled in the server configuration, or;
    3. If the underlying transport does not support unreliable datagrams (HTTP/1.1 and HTTP/2).

### Result

Support for reading/writing unreliable datagrams introduced.


Sources/NIOHTTPServer/Configuration/HTTP3/HTTP3+ConnectionSettings.swift
```diff
@@ -37,12 +37,31 @@ extension NIOHTTPServerConfiguration.HTTP3 {
         ///   `SETTINGS_MAX_FIELD_SECTION_SIZE`.
         public var maximumFieldSectionSize: UInt64?
 
+        /// Whether the server should advertise support for receiving HTTP/3 datagrams.
+        ///
+        /// - SeeAlso: https://www.rfc-editor.org/rfc/rfc9297.html#section-2.1.1-1. Corresponds to
+        ///   `SETTINGS_H3_DATAGRAM`.
+        var http3Datagram: Bool
```
gjcairo: I'd name this something like `areHTTP3DatagramsSupported` or `supportHTTP3Datagrams` to make it clear it's a boolean.

Also, where are we wiring this up?

Sources/NIOHTTPServer/Configuration/HTTP3/HTTP3+QUICConfiguration.swift
```diff
@@ -159,6 +159,43 @@ extension NIOHTTPServerConfiguration.HTTP3 {
         /// debugging and analysis.
         public var qLogConfiguration: QLogConfiguration?
 
+        /// The maximum datagram frame size in bytes. If set to 0, the server will not advertise support for receiving
+        /// unreliable datagrams.
+        var maxDatagramFrameSize: Int
```
gjcairo: Should this not be public?

Sources/NIOHTTPServer/Configuration/HTTP3/HTTP3+QUICConfiguration.swift
```diff
@@ -159,6 +159,43 @@ extension NIOHTTPServerConfiguration.HTTP3 {
...
+
+        init(
```
gjcairo: Do we actually need this init?

Sources/NIOHTTPServer/Configuration/HTTP3/HTTP3+QUICConfiguration.swift
gjcairo: Do we actually need this in the QUIC configuration or is it something NIOHTTP3 may set if datagrams are enabled at the H3 level?

Sources/NIOHTTPServer/Documentation.docc/SwiftConfigurationIntegration.md
```diff
@@ -70,6 +70,9 @@ its respective key prefix.
 | `http.http3.quicConfiguration.qlog` | `path`                                    | `string`        | Optional                                                                                                                      | nil            |
 |                                     | `topic`                                   | `string`        | Optional                                                                                                                      | nil            |
 |                                     | `description`                             | `string`        | Optional                                                                                                                      | nil            |
+| `http.http3.datagramConfiguration`  | `datagramsEnabled`                        | `bool`          | Optional                                                                                                                      | true           |
```
gjcairo: Technically, this will only be enabled (and exposed) only if the corresponding unstable trait is enabled. Should we make this explicit in the docs?

Tests/NIOHTTPServerTests/NIOHTTPServerWriterTests.swift
```diff
@@ -118,10 +123,13 @@ struct NIOHTTPServerWriterTests {
     @available(anyAppleOS 26.0, *)
     func takeDatagramWriterVendsResponseAndDatagramWriter() async throws {
         let (outboundWriter, sink) = NIOAsyncChannelOutboundWriter<HTTPResponsePart>.makeTestingWriter()
+        let connectionChannel = EmbeddedChannel()
         let sender = NIOHTTPServer.ResponseSender(
             writer: outboundWriter,
             writerState: .init(),
-            datagramWriter: NIOHTTPServer.DatagramWriter()
+            datagramWriter: NIOHTTPServer.DatagramWriter(
+                unreliableStream: .init(streamID: 4, connectionChannel: connectionChannel, maxBufferedDatagrams: 16)
```
gjcairo: I know the H3 pieces aren't fully there yet but we should add a stub for a test that checks the max number of buffered datagrams is respected.

Sources/NIOHTTPServer/Datagrams/HTTP3DatagramDemultiplexer.swift
```diff
@@ -0,0 +1,56 @@
...
+@available(anyAppleOS 26.0, *)
+final class HTTP3DatagramDemultiplexer: ChannelInboundHandler {
+    typealias InboundIn = HTTP3Datagram
+
+    /// The ``HTTP3DatagramStream`` instance for each open request stream.
+    private var datagramStreams: [QUICStreamID: HTTP3UnreliableDatagramStream] = [:]
+
+    /// Starts routing datagrams received for `datagramStream.streamID` to the provided `datagramStream`.
+    ///
+    /// - Precondition: Must only be called on the connection channel's event loop.
```
gjcairo: We should actually precondition. Same in the func below.

Sources/NIOHTTPServer/Datagrams/HTTP3UnreliableDatagramStream.swift
```diff
@@ -0,0 +1,74 @@
...
+        self.maxBufferedDatagrams = maxBufferedDatagrams
+
+        (self.inbound, self.inboundContinuation) = AsyncStream<ByteBuffer>.makeStream(
+            bufferingPolicy: .bufferingNewest(maxBufferedDatagrams)
+        )
+    }
+
+    /// Delivers an inbound datagram to the request stream's datagram reader.
+    ///
+    /// - Precondition: Must be called on the connection channel's event loop.
```
gjcairo: I don't actually think this is true. Same below.

Sources/NIOHTTPServer/NIOHTTPServer+HTTP3.swift
```diff
@@ -218,20 +214,44 @@ extension NIOHTTPServer {
...
-        NIOQUIC.QUICStreamCreator
-    > {
+    ) throws -> HTTP3ServerConnection<HTTP3Stream, NIOQUIC.QUICStreamCreator> {
         let loopBoundHandler = NIOLoopBoundBox<HTTP3ConnectionHandler<NIOQUIC.QUICStreamCreator>?>(
             nil,
             eventLoop: connectionChannel.eventLoop
         )
 
+        #if UnstableHTTPDatagrams
+        // TODO: If support for datagrams was not negotiated, we shouldn't create the datagram handler and demultiplexer.
```
gjcairo: We can also skip it if the datagrams trait is disabled no?

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
```diff
@@ -366,6 +366,64 @@ extension NIOHTTPServer {
...
+                error: error,
+                metadata: [LoggingKeys.protocol: "\(context.httpVersion)"]
+            )
+            try? await channel.channel.close()
+        }
+    }
+    #endif  // HTTP3 && UnstableHTTPDatagrams
+
     /// Handles a stream channel, which carries exactly one request per stream.
     func handleStreamChannel<Handler: HTTPServerRequestHandler>(
```
gjcairo: Should we have this second version in an `#else` for the `UnstsableHTTPDatagrams` trait?

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -345,6 +345,71 @@ public struct NIOHTTPServer: HTTPServer {
...
+            if !writerState.wrapped.withLock({ $0.finishedWriting }) {
+                Self.abortRequest(requestContext: requestContext, error: error)
+            }
+        }
+
+        if !writerState.wrapped.withLock({ $0.finishedWriting }) {
+            self.logger.debug("Handler did not conclude the response.")
+        }
+    }
+    #endif
```
gjcairo: Same here: should we have the version of `invokeHandler` without the datagram stream in the `#else` branch?

Tests/NIOHTTPServerTests/HTTP3DatagramTests.swift
```diff
@@ -0,0 +1,196 @@
...
+import NIOCore
+import NIOEmbedded
+import NIOHTTP3
+import NIOQUICHelpers
+import Testing
+
+@testable import NIOHTTPServer
+
+@Suite
+struct HTTP3DatagramTests {
```
gjcairo, line 29: Can we add a test for checking that if a write fails because the datagram size is larger than the configured limit, subsequent writes succeed?

gjcairo (changes requested):

Sources/NIOHTTPServer/Configuration/HTTP3/HTTP3+QUICConfiguration.swift
aryan-25: To advertise support for receiving datagrams:
1. The QUIC layer has to send a `max_datagram_frame_size` QUIC transport parameter with value > 0, and;
2. The HTTP3 layer has to send `SETTINGS_H3_DATAGRAMS` in the SETTINGS frame.

We need to set this config in both places; the QUIC connection handshake has already taken place before NIOHTTP3 starts.

Sources/NIOHTTPServer/NIOHTTPServer+HTTP3.swift
aryan-25: Yes, this is already the case.

Sources/NIOHTTPServer/Configuration/HTTP3/HTTP3+ConnectionSettings.swift
aryan-25: Yes, agree that the naming should be better. We should also fix this in swift-nio-http3. 

This is being wired up [here](https://github.com/aryan-25/swift-http-server/blob/e9ba477c6506457f20ab962a31dc57f7f2069efa/Sources/NIOHTTPServer/Configuration/HTTP3/HTTP3%2BConnectionSettings.swift#L82) and [here](https://github.com/aryan-25/swift-http-server/blob/e9ba477c6506457f20ab962a31dc57f7f2069efa/Sources/NIOHTTPServer/NIOHTTPServer%2BHTTP3.swift#L277).

Sources/NIOHTTPServer/Datagrams/HTTP3DatagramDemultiplexer.swift
aryan-25: Addressed in commit [6426ec1](https://github.com/swift-server/swift-http-server/pull/117/commits/6426ec1643c2bb2a9218c79664c0d47438de78ca).

Sources/NIOHTTPServer/Datagrams/HTTP3UnreliableDatagramStream.swift
aryan-25: Addressed in commit [42d6d0a](https://github.com/swift-server/swift-http-server/pull/117/commits/42d6d0a9c1318ce629a5d3cb21bcf88775032e01).

Sources/NIOHTTPServer/Documentation.docc/SwiftConfigurationIntegration.md
aryan-25: Addressed in commit [2139711](https://github.com/swift-server/swift-http-server/pull/117/commits/2139711d0f63d54c7ea9eb45b1fe56a40aac504b).

Sources/NIOHTTPServer/NIOHTTPServer+HTTP3.swift
gjcairo: Ah, is the "datagram handler" you mention in this TODO the `HTTP3UnreliableDatagramStream` from below?

aryan-25: Sorry, that was poorly worded. What I meant by "datagram handler" was the stub handler. But the stub is gone now since the support in swift-nio-http3 landed.

Sources/NIOHTTPServer/Configuration/HTTP3/HTTP3+QUICConfiguration.swift
aryan-25: I've cleaned this up in commit [faf6499](https://github.com/swift-server/swift-http-server/pull/117/commits/faf649959814126ce02790d03e315ed874e44a8f).

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
aryan-25: Addressed in commit [2853f0a](https://github.com/swift-server/swift-http-server/pull/117/commits/2853f0ab15c2a3e5f7a039474637660111b0b649).

Sources/NIOHTTPServer/NIOHTTPServer.swift
aryan-25: See https://github.com/swift-server/swift-http-server/pull/117#discussion_r3874206504.

Tests/NIOHTTPServerTests/HTTP3DatagramTests.swift
aryan-25: Addressed in commit [98817ba](https://github.com/swift-server/swift-http-server/pull/117/commits/98817ba38687eca916842e861f187db34109a38a).

Tests/NIOHTTPServerTests/NIOHTTPServerWriterTests.swift
aryan-25: Addressed [here](https://github.com/aryan-25/swift-http-server/blob/b872fd65d23e9970eb60ff342885ace1fd059a82/Tests/NIOHTTPServerTests/NIOHTTPServerReaderTests.swift#L295).

Tests/NIOHTTPServerTests/HTTP3DatagramTests.swift
```diff
@@ -270,5 +271,82 @@ struct HTTP3DatagramTests {
...
+                // Try to write a datagram larger than 150 bytes.
+                var writer = try await responseSender.send(.init(status: .ok))
+                var datagramWriter = await writer.takeDatagramWriter()!
+
+                let largerThanMaxFrameSize = Array<UInt8>(repeating: 1, count: clientMaxDatagramFrameSize + 50)
+                var bufferOne = UniqueArray<UInt8>(copying: largerThanMaxFrameSize.span)
+
+                await #expect(throws: QUICError.datagramTooLarge) {
+                    try await datagramWriter.write(buffer: &bufferOne)
+                }
```
gjcairo: Didn't we speak offline with @josephnoir who said that QUIC drops datagrams that are too large? Or is that in a different scenario?

aryan-25, line 322: Yeah, there are two scenarios:
1. If we try to write a datagram larger than the `max_datagram_frame_size` advertised by the peer, `swift-nio-quic` reports an error to us (`QUICError.datagramTooLarge`).
2. If we try to write a datagram larger than the MSS, then the datagram gets dropped by `swift-nio-quic` without us knowing. See [this](https://github.com/apple/swift-network-evolution/blob/59ba0e49b5bab8c4d0b320608a5474f5da72e0b2/Sources/SwiftNetwork/QUIC/SendItems.swift#L1600-L1606).

Please correct me if I'm wrong @josephnoir.

This test just checks if we get the error reported to us in scenario (1).

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
```diff
@@ -365,56 +376,71 @@ extension NIOHTTPServer {
...
-        Handler.RequestContext == RequestContext,
-        Handler.Reader == Reader,
-        Handler.ResponseSender == ResponseSender
-    {
+extension NIOAsyncChannelInboundStream<HTTPRequestPart>.AsyncIterator {
+    /// Reads the next request head from the iterator. Returns `nil` if the connection is done or an unexpected part is
+    /// received.
+    ///
+    /// Skips over leftover `.body` and `.end` parts from a previous request that the handler didn't fully consume.
+    mutating func nextRequestHead(logger: Logger) async throws -> HTTPRequest? {
```
gjcairo: Why did we move this here from `NIOHTTPServer`?

Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
```diff
@@ -182,3 +162,26 @@ extension NIOHTTPServer.ResponseSender: Sendable {}
 
 @available(*, unavailable)
 extension NIOHTTPServer.ResponseSender.Writer: Sendable {}
+
+extension ByteBuffer {
+    /// Drains `buffer` into a newly allocated `ByteBuffer`.
+    init<Buffer: RangeReplaceableContainer<UInt8> & ~Copyable>(
```
gjcairo, line 168: Don't we have this elsewhere already?

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -416,6 +383,67 @@ public struct NIOHTTPServer: HTTPServer {
...
+            request: request,
+            requestContext: requestContext,
+            requestReader: requestReader,
+            responseSender: responseSender,
+            handler: handler
+        )
+    }
+    #endif  // HTTP3 && UnstableHTTPDatagrams
+
+    func invokeHandler<Handler: HTTPServerRequestHandler>(
```
gjcairo, line 420: I don't really like the fact we've got overloads for the `invokeHandler` method. Why did we break this up exactly? Can we name them something else? Also for these two variants, can we keep the one without H3 in an `#else` branch?

Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
aryan-25: No, we were repeating this logic twice in this file and we also need it to implement the datagram writer. I thought of putting it into a helper so we can avoid the duplication.

Tests/NIOHTTPServerTests/Utilities/TestClientConnection.swift
```diff
@@ -149,10 +156,78 @@ extension TestClientConnection {
...
 @available(anyAppleOS 26.0, *)
 extension TestClientConnection {
+    /// The information needed to establish a test client connection to a ``NIOHTTPServer``.
+    struct Configuration {
+        var logger: Logger
+        var httpVersion: Version
+        var trustRootsPEMPath: String?
+        var chain: ChainPrivateKeyPair? = nil
+
+        enum Version {
```
gjcairo: Do we really need yet another version enum?

Sources/NIOHTTPServer/NIOHTTPServer.swift
aryan-25: > Why did we break this up exactly?

We need to pass the datagramStreamFuture to the request reader and response writer if HTTP/3 is among the supported versions and datagrams are enabled. But the datagram types are behind the `UnstableHTTPDatagrams` trait, so we need two variants of this method.

I've renamed the datagram variant to `invokeDatagramEnabledHandler`.

Sources/NIOHTTPServer/NIOHTTPServer+SecureUpgrade.swift
aryan-25: Sorry, forgot to clean this up when I was refactoring that method. I've moved it back to `NIOHTTPServer` now.

Note that this method is now an extension on `NIOAsyncChannel`.

Tests/NIOHTTPServerTests/Utilities/TestClientConnection.swift
aryan-25: Addressed in commit [e9fed89](https://github.com/swift-server/swift-http-server/pull/117/commits/e9fed89c95dbafd5e84e44a2f15f79b0dc6513d0).

Sources/NIOHTTPServer/NIOHTTPServer+HTTP3.swift
```diff
@@ -106,23 +106,26 @@ extension NIOHTTPServer {
...
+                                inboundIterator: inboundIterator,
+                                outbound: outbound,
+                                datagramStreamFuture: datagramStreamFuture,
+                                handler: handler
+                            )
+                            return
+                        }
+                        #endif  // UnstableHTTPDatagrams
+
                         _ = await self.invokeHandler(
```
gjcairo, line 122: Do we need to return something from this method btw?

aryan-25: We don't need to return the iterator for H2 and H3.

gjcairo (approved):

PR #124, opened 2026-08-27
Add RawSpan APIs to the ResponseSender

### LINKED ISSUE #123: Offer ResponseSender.Writer RawSpan APIs
We should offer `RawSpan` APIs for the ResponseSender.Writer to avoid an extra copy when it takes the container and copies it to a ByteBuffer. If we have a `RawSpan` API we can bypass that allocation for every write. I verified this in Vapor 5 where currently we have to stage every chunk into a `UniqueArray`. I've tested this and it reduces the copy for every write, which can add up when streaming large responses. It reduces it down to a single copy in the pipeline. Will submit a PR with the new APIs.

We can write overloads to do a borrow instead of the copy:

```swift
public mutating func write(_ bytes: RawSpan) async throws(WriteFailure)
public consuming func finish(bytes: RawSpan, finalElement: consuming HTTPFields?) async throws(WriteFailure)
```

Note that we'd need an overload for `finish` as well which covers non-streaming cases.

0xTim:
Resolves #123

### DIFF
--- a/Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
+++ b/Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
@@ -137,6 +137,40 @@ extension NIOHTTPServer.ResponseSender {
             try await self.writer.write(.body(byteBuffer))
         }
 
+        /// Writes a chunk of the response body from borrowed memory.
+        ///
+        /// - Parameter bytes: The body bytes to write. Only borrowed for the duration of the call.
+        public mutating func write(_ bytes: RawSpan) async throws(WriteFailure) {
+            var byteBuffer = ByteBuffer()
+            byteBuffer.reserveCapacity(bytes.byteCount)
+            byteBuffer.writeBytes(bytes)
+
+            try await self.writer.write(.body(byteBuffer))
+        }
+
+        /// Writes a final chunk of the response body from borrowed memory and completes the response.
+        ///
+        /// The borrowing counterpart to ``finish(buffer:finalElement:)``
+        ///
+        /// - Parameters:
+        ///   - bytes: The final body bytes. Only borrowed for the duration of the call. May be empty,
+        ///     in which case no body part is written.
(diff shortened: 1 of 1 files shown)


Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
```diff
@@ -137,6 +137,40 @@ extension NIOHTTPServer.ResponseSender {
             try await self.writer.write(.body(byteBuffer))
         }
 
+        /// Writes a chunk of the response body from borrowed memory.
+        ///
+        /// - Parameter bytes: The body bytes to write. Only borrowed for the duration of the call.
+        public mutating func write(_ bytes: RawSpan) async throws(WriteFailure) {
+            var byteBuffer = ByteBuffer()
+            byteBuffer.reserveCapacity(bytes.byteCount)
+            byteBuffer.writeBytes(bytes)
```
FranzBusch, line 146: The byte buffer allocation here makes this path really sad.

0xTim: It needs to be copied into a ByteBuffer at _some_ point because that's what the channel handler takes. The HTTP server needs to own the data because of the difference of when NIO returns and when the data actually hits the socket (IIRC this is a documented part of the channel handler). This APIs here at least stops you going from Data/Array<UInt8>/ByteBuffer/whatever type you want, into a unique array back into ByteBuffer in the channel handler

FranzBusch: The only reason we need this API is because `ByteBuffer` does not conform to `RangeReplacableContainer` nor do the span types.

0xTim: We'd still have a copy somewhere, unless it was ByteBuffer through the whole stack right? (Assuming `RangeReplaceableContainer`). Because that's what NIO talks in the channel handler/socket so it needs a copy to convert it in. I don't think `ByteBuffer` has any mechanism to borrow yet

PR #125, opened 2026-08-29
Expose the negotiated HTTP version to request handlers

### LINKED ISSUE #113: Expose the negotiated HTTP version to request handlers
### What I'm trying to do

Read, inside a request handler, the application protocol negotiated for the
connection (HTTP/1.1 vs HTTP/2, etc.).

### The gap

`NIOHTTPServer.ConnectionContext` already carries it as a `public let httpVersion: HTTPVersion`, but a handler can't reach it: the only thing it receives is a `RequestContext`, whose `connectionContext` is `internal`. `RequestContext` surfaces the peer/local addresses (`ConnectionInfo`) and the validated peer certificate chain (`PeerCertificate`), but there's no capability for the negotiated version — so a handler has no way to tell which protocol it's serving.

### Suggested shape

Expose it following the existing capability pattern, e.g.:

```
    extension HTTPServerCapability {
        public protocol NegotiatedHTTPVersion: RequestContext {
            var httpVersion: NIOHTTPServer.HTTPVersion { get }
        }
    }

    extension NIOHTTPServer.RequestContext: HTTPServerCapability.NegotiatedHTTPVersion {
        public var httpVersion: NIOHTTPServer.HTTPVersion {
            self.connectionContext.httpVersion
        }
    }
```

This keeps `ConnectionContext` internal to the handler surface and stays consistent with how `ConnectionInfo` / `PeerCertificate` are exposed. A minimal alternative would be just the `public var httpVersion` accessor. Happy to open a PR for whichever shape you prefer.

dioKaratzas:
Resolves #113.

Adds `HTTPServerCapability.NegotiatedHTTPVersion`, following the existing `ConnectionInfo` / `PeerCertificate` pattern, and conforms `NIOHTTPServer.RequestContext` to it by forwarding to the connection context. `ConnectionContext` stays internal to the handler surface. Purely additive.

Test plan: asserted the negotiated version from the request handler in the existing HTTP/1.1 and HTTP/2 end-to-end tests.

### DIFF
--- a/Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
+++ b/Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
@@ -31,6 +31,16 @@ extension HTTPServerCapability {
         var localAddress: NIOHTTPServer.SocketAddress? { get }
     }
 
+    /// A request-context capability exposing the negotiated application protocol.
+    ///
+    /// Servers whose request context conforms to this capability surface the
+    /// application-level HTTP version negotiated for the connection carrying
+    /// the request.
+    public protocol NegotiatedHTTPVersion: RequestContext {
+        /// The application-level HTTP version negotiated for the connection.
+        var httpVersion: NIOHTTPServer.HTTPVersion { get }
+    }
+
     /// A request-context capability exposing the validated peer certificate chain.
     ///
     /// Servers whose request context conforms to this capability surface the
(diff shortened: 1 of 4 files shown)


gjcairo (approved):

gjcairo:
Thanks  @dioKaratzas !

Sources/NIOHTTPServer/HTTPServerCapability+ConnectionCapabilities.swift
```diff
@@ -31,6 +31,16 @@ extension HTTPServerCapability {
         var localAddress: NIOHTTPServer.SocketAddress? { get }
     }
 
+    /// A request-context capability exposing the negotiated application protocol.
+    ///
+    /// Servers whose request context conforms to this capability surface the
+    /// application-level HTTP version negotiated for the connection carrying
+    /// the request.
+    public protocol NegotiatedHTTPVersion: RequestContext {
```
FranzBusch, line 39: I think `Negotiated` is potentially wrong here. If the server runs over bare TCP without TLS then nothing is negotiated necessarily. What about just `HTTPVersion`?

PR #132, opened 2026-09-09
Improve handling of socket lifecycle

aryan-25:
### Motivation

We currently return bound listening sockets in `NIOHTTPServer.makeServerChannels` and start iterating over incoming requests on those sockets in `NIOHTTPServer.serve`. However, bound sockets should not be returned from a method. The lifecycle of a socket should be managed more carefully.

### Modifications

- Removed methods that returned bound channels and replaced them with `with`-style methods.
- Removed `ServerChannel` and replaced it with `ListenerConfiguration`, which now describes what sockets should be setup.
- Refactored the logic in `NIOHTTPServer.serve`. It now iterates over each bind target and starts a task which both creates the socket and serves requests from the socket.
  - With this, we can now also set the task executor preference to the socket's event loop.
- Some general refactors. 

### Result

Better handling of socket lifecycle.


Sources/NIOHTTPServer/ListenerConfiguration.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    #if HTTP3
+    struct HTTP3: Sendable {
+        let http3Configuration: NIOHTTPServerConfiguration.HTTP3
+        let authenticationConfiguration: NIOQUIC.AuthenticationConfiguration
+        let quicAuthenticator: NIOQUIC.Authenticator?
+    }
+    #endif
+
+    case plaintextHTTP1_1
```
fabianfett, line 37: not something to fix in this pr, but will we support `plaintextHTTP2`?

aryan-25: Yes, this is planned.

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -177,24 +182,86 @@ public struct NIOHTTPServer: HTTPServer {
...
+                case .plaintextHTTP1_1:
+                    self.addPlaintextHTTP1_1Listener(
+                        to: &group,
+                        address: try NIOCore.SocketAddress(bindTarget: bindTarget),
+                        addressContinuation: addressContinuation,
+                        connectionHandler: connectionHandler
+                    )
+
+                    let resolvedAddress = try await self.nextBoundAddress(from: &addressStreamIterator)
+                    boundAddresses.append(resolvedAddress)
```
fabianfett: can we move this just below the switch? We repeat it 4x right now

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -622,3 +542,105 @@ extension NIOAsyncChannelInboundStream<HTTPRequestPart>.AsyncIterator {
...
+        let (bootstrap, serverQuiescingHelper) = ServerBootstrap.makeHTTPServerBootstrap(
+            group: self.eventLoopGroup,
+            maxConnections: self.configuration.maxConnections
+        )
+
+        let serverChannel: NIOAsyncChannel<Child, Never>
+        do {
+            serverChannel = try await bootstrap.bind(to: address) { channel in
+                childChannelInitializer(channel)
+            }
```
fabianfett: can you pass the childChannelInitializer directly?
something like?
```
            serverChannel = try await bootstrap.bind(to: address, initializer: childChannelInitializer)
```

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -622,3 +542,105 @@ extension NIOAsyncChannelInboundStream<HTTPRequestPart>.AsyncIterator {
         }
     }
 }
+
+@available(anyAppleOS 26.0, *)
+extension ServerBootstrap {
+    /// Makes a `ServerBootstrap` alongside the `ServerQuiescingHelper` used to later shut that listener down gracefully.
+    ///
+    /// - Note: A `ConnectionLimitHandler` is only installed when `maxConnections` is non-`nil`.
+    static func makeHTTPServerBootstrap(
```
fabianfett: rename to makeTCPServerBootstrap is likely better, there is nothing HTTP about this one

Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
```diff
@@ -111,66 +111,33 @@ extension NIOHTTPServer {
...
-                        )
-                    }
-                    serverChannels.append((serverChannel, serverQuiescingHelper))
+    /// Adds a child task to `group` that binds a plaintext HTTP/1.1 listener at `address` and serves connections on it
+    /// until the task is cancelled or the server shuts down gracefully.
+    ///
+    /// - Note: The bind address is yielded to the provided `addressContinuation` immediately after the TCP socket has
+    ///   been bound.
+    func addPlaintextHTTP1_1Listener<Handler: NIOHTTPServerConnectionHandler>(
+        to group: inout ThrowingDiscardingTaskGroup<any Error>,
```
FranzBusch, line 118: Why is the task group throwing? What happens if one of the child tasks throws? Does it bring down the entire server?

Sources/NIOHTTPServer/NIOHTTPServer+HTTP3.swift
```diff
@@ -132,73 +134,114 @@ extension NIOHTTPServer {
...
+    ///   been bound.
+    func addHTTP3Listener<Handler: NIOHTTPServerConnectionHandler>(
+        to group: inout ThrowingDiscardingTaskGroup<any Error>,
+        address: NIOCore.SocketAddress,
+        eventLoop: any EventLoop,
+        configuration: ListenerConfiguration.HTTP3,
+        addressContinuation: AsyncThrowingStream<NIOCore.SocketAddress, any Error>.Continuation,
+        connectionHandler: Handler
+    ) {
+        let eventLoopExecutor = eventLoop.executor as? any TaskExecutor
```
FranzBusch, line 150: EventLoops are not conforming to `TaskExecutor` yet so this presumably just returns `nil`?

Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
```diff
@@ -111,66 +111,33 @@ extension NIOHTTPServer {
...
+        address: NIOCore.SocketAddress,
+        addressContinuation: AsyncThrowingStream<NIOCore.SocketAddress, any Error>.Continuation,
+        connectionHandler: Handler
+    ) {
+        group.addTask(name: "Plaintext HTTP/1.1 over \(address)") {
```
FranzBusch, line 123: Why are we not using task executor preference here?

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -622,3 +542,105 @@ extension NIOAsyncChannelInboundStream<HTTPRequestPart>.AsyncIterator {
...
+                addressContinuation.yield(localAddress)
+
+                do {
+                    try await body(serverChannel)
+                } catch {
+                    try? await serverChannel.executeThenClose { _ in }
+                    throw error
+                }
+
+                try? await serverChannel.executeThenClose { _ in }
```
fabianfett: I don't know enough about NIOAsyncChannel, but should the bode be executed in the `executeThenClose` closure?

fabianfett (commented):
I love all of this. Few small questions. I think @gjcairo should also have a look.

Sources/NIOHTTPServer/NIOHTTPServer+HTTP3.swift
```diff
@@ -132,73 +134,114 @@ extension NIOHTTPServer {
...
+            addressContinuation.finish(throwing: error)
             throw error
         }
 
-        return serverChannels
+        try await withTaskCancellationHandler {
+            try await withGracefulShutdownHandler {
+                if Task.isCancelled || Task.isShuttingDownGracefully {
+                    // The cancellation/shutdown handler will have cleaned up the socket. Just return here.
+                    return
```
FranzBusch, line 200: This is not enough. All paths through this method must await the `channel.closeFuture` only then is this with-style method safe and ensures that every resource is cleaned up.

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -177,24 +182,86 @@ public struct NIOHTTPServer: HTTPServer {
...
         defer { self.finishListeningAddressPromise() }
 
-        let serverChannels = try await self.makeServerChannels()
-
-        return try await withTaskCancellationHandler {
-            try await withGracefulShutdownHandler {
-                try await self._serve(serverChannels: serverChannels, connectionHandler: connectionHandler)
-            } onGracefulShutdown: {
-                self.beginGracefulShutdown(serverChannels: serverChannels)
+        try await withThrowingDiscardingTaskGroup { group in
```
FranzBusch, line 190: I think this group should be non-throwing

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -622,3 +542,105 @@ extension NIOAsyncChannelInboundStream<HTTPRequestPart>.AsyncIterator {
...
+        } catch {
+            addressContinuation.finish(throwing: error)
+            throw error
+        }
+
+        try await withTaskCancellationHandler {
+            try await withGracefulShutdownHandler {
+                if Task.isCancelled || Task.isShuttingDownGracefully {
+                    // The cancellation/shutdown handler will have cleaned up the socket. Just return here.
+                    return
```
FranzBusch, line 618: Same here. You have to await the close future.

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -622,3 +542,105 @@ extension NIOAsyncChannelInboundStream<HTTPRequestPart>.AsyncIterator {
...
+                    addressContinuation.finish(throwing: ListeningAddressError.addressOrPortNotAvailable)
+                    throw ListeningAddressError.addressOrPortNotAvailable
+                }
+
+                addressContinuation.yield(localAddress)
+
+                do {
+                    try await body(serverChannel)
+                } catch {
+                    try? await serverChannel.executeThenClose { _ in }
```
FranzBusch: I think you should just call `serverChannel.channel.close()`

FranzBusch (commented):
This is overall good except a few minor comments. Some of this should really be API in NIO since it is generally useful

Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
aryan-25: > What happens if one of the child tasks throws? Does it bring down the entire server?

Correct. Note that the child tasks will _not_ throw if there's a connection or stream error (see [this](https://github.com/swift-server/swift-http-server/blob/aa39e5c6c1934a2a06b26d2977c67bbc3846ab8b/Sources/NIOHTTPServer/NIOHTTPServer%2BHTTP1_1.swift#L41-L44)). The child tasks only throw if there's an error at the main listening socket level, e.g. if the listening socket couldn't be bound.

In that case, we want to clean up all previously bound sockets and bring down the whole server. We had this property before this PR.

Sources/NIOHTTPServer/NIOHTTPServer+HTTP3.swift
aryan-25: Correct. We need https://github.com/apple/swift-nio/pull/2732 to land.

Sources/NIOHTTPServer/NIOHTTPServer.swift
aryan-25: The listener tasks need to throw if there's an error at the listening socket level, as discussed [here](https://github.com/swift-server/swift-http-server/pull/132#discussion_r3989027981).

Sources/NIOHTTPServer/NIOHTTPServer+HTTP1_1.swift
fabianfett: I think we should do a follow up for this one. we need task executor preference for the listening socket and then a different one per accepted connection. the mapping is not as easy as http3.

Sources/NIOHTTPServer/NIOHTTPServer.swift
aryan-25: Addressed in commit [cd7b546](https://github.com/swift-server/swift-http-server/pull/132/commits/cd7b546557421d3ce10ac93169032e8c073fb2a7).

Sources/NIOHTTPServer/NIOHTTPServer.swift
aryan-25: I've reduced some duplication in commit [d98b820](https://github.com/swift-server/swift-http-server/pull/132/commits/d98b820e2916e9d4017d4a74a6275ddb9576629c). But unfortunately, we need to wait for the resolved address in each of the four cases, because in the `secureUpgradeAndHTTP3` case, we use the TCP socket's resolved address as the bind address for the UDP socket.

Sources/NIOHTTPServer/NIOHTTPServer.swift
aryan-25: Addressed in commit [8d22ac3](https://github.com/swift-server/swift-http-server/pull/132/changes/8d22ac330346edf52a864747048de03dc1240bca).

Sources/NIOHTTPServer/NIOHTTPServer.swift
aryan-25: Thanks for this. I realised we are just duplicating the logic of `executeThenClose` here. I've fixed this in commit [9cff2f4](https://github.com/swift-server/swift-http-server/pull/132/commits/9cff2f4f8224bb069cca7d02bffd8f5deef31e8f).

Sources/NIOHTTPServer/NIOHTTPServer+HTTP3.swift
aryan-25: Addressed in commit [6015ac0](https://github.com/swift-server/swift-http-server/pull/132/commits/6015ac09959a5e8c1811fcd8ac23358dc8acb709).

Sources/NIOHTTPServer/NIOHTTPServer.swift
aryan-25: Fixed. See https://github.com/swift-server/swift-http-server/pull/132#discussion_r3990248876.

Sources/NIOHTTPServer/NIOHTTPServer+HTTP3.swift
FranzBusch: I would prefer if we keep the task executor preference setting here out of the PR until then and follow up with a holistic application of it.

aryan-25: Sounds good. I've removed it in commit [64caf14](https://github.com/swift-server/swift-http-server/pull/132/commits/64caf1480f69f42d3e8cf32a639bf4c04814da8f).

FranzBusch (approved):

gjcairo (approved):

fabianfett (approved):

PR #134, opened 2026-09-11
Cancel request handler when client goes away

### LINKED ISSUE #53: Client closing the socket does not cause a CancellationError on the server
We have found that a server that is async-waiting after receiving a HTTP request does not see a `CancellationError` if the client were to close the connection.

This is trivially reproducible.

Server code:
```
let logger = Logger(label: "TestHTTPServer")
let server = NIOHTTPServer(logger: logger, configuration: .init(bindTarget: .hostAndPort(host: "127.0.0.1", port: 12345)))
try await server.serve { request, requestContext, requestBodyAndTrailers, responseSender in
    print("STALLING...")
    try await Task.sleep(for: .seconds(60 * 60))
}
```

Client terminal:
```
$ curl -v http://localhost:12345/stall
* Host localhost:12345 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:12345...
* connect to ::1 port 12345 from ::1 port 57145 failed: Connection refused
*   Trying 127.0.0.1:12345...
* Connected to localhost (127.0.0.1) port 12345
> GET /stall HTTP/1.1
> Host: localhost:12345
> User-Agent: curl/8.7.1
> Accept: */*
> 
* Request completely sent off
^C
```

Server output:
```
STALLING...
```

The server hangs even though the client has closed the socket. Note how there is no exception or additional printing in the server output.

gjcairo:
This PR adds a mechanism to automatically cancel the request handler whenever a client goes away.

- For H1, this means when the connection is closed.
- For H2, when the stream is reset or the underlying connection closes.
- For H3, when both STOP_SENDING and RESET_STREAM are exchanged, or when the underlying connection closes.


Sources/NIOHTTPServer/ClientClosed.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    private let clientClosed: AsyncStream<Void>.Continuation
+
+    init(clientClosed: AsyncStream<Void>.Continuation) {
+        self.clientClosed = clientClosed
+    }
+
+    // - TODO: HTTP/1.1 on Darwin does not report a socket closing, as NIO registers for EOF
+    // notifications only when `isEarlyEOFDeliveryWorkingOnThisOS` is true, which is hard-coded `false`
+    // on Darwin (rdar://53656794). This means that this handler has no effect on H1 on Darwin until
```
FranzBusch: Please no radar links here.

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -127,6 +127,22 @@ public struct NIOHTTPServer: HTTPServer {
     /// Throwing after the response has been concluded aborts nothing: a complete response is never retracted, so the
     /// only consequence is that the connection is not reused.
     ///
+    /// ## Cancellation when the exchange ends
+    ///
+    /// The task running a handler is cancelled if the exchange carrying its request ends before the handler returns,
```
FranzBusch, line 132: exchange is kind of a weird word to use. What about transport?

Sources/NIOHTTPServer/ClientClosed.swift
```diff
@@ -0,0 +1,62 @@
...
+
+/// Runs `operation` and cancels it if the client stops waiting for the response first.
+///
+/// - Note: `operation` is a whole HTTP/1.1 request loop or a whole HTTP/2 or HTTP/3 stream.
+func withCancellationWhenClientCloses(
+    signalledBy clientClosed: AsyncStream<Void>,
+    operation: @escaping @Sendable () async throws -> Void
+) async throws {
+    try await withThrowingTaskGroup(of: Void.self) { group in
+        group.addTask {
```
FranzBusch, line 49: Bit unfortunate that we need to child tasks here but okay for now

Sources/NIOHTTPServer/ClientClosed.swift
gjcairo: Eek sorry, I thought I'd removed this.

Sources/NIOHTTPServer/ClientClosed.swift
gjcairo: Any alternatives I'm missing?

Sources/NIOHTTPServer/NIOHTTPServer.swift
gjcairo: Hmm if this was H2 or H3 only I'd use "stream or connection", if H1 I'd use "connection". I was trying to find a word that encapsulated both but I agree exchange isn't it. Maybe just "the stream or connection"?

Sources/NIOHTTPServer/NIOHTTPServer.swift
```diff
@@ -127,6 +127,22 @@ public struct NIOHTTPServer: HTTPServer {
...
+    /// so a handler awaiting work nobody will read stops instead of running to completion. Handlers that hold
+    /// resources or drive requests of their own should therefore honour cancellation.
+    ///
+    /// What counts as the exchange ending depends on the protocol:
+    ///
+    /// - Over HTTP/1.1, the connection closing.
+    /// - Over HTTP/2, the stream closing, whether from the client's `RST_STREAM` or from the connection beneath it
+    ///   going away.
+    /// - Over HTTP/3, the stream closing, or receiving **both** `STOP_SENDING` and `RESET_STREAM` while it is still
```
ehaydenr, line 141: What does "the stream closing" mean as opposed to receiving both STOP_SENDING and RESET_STREAM? If stream is only closed once either fin is sent in each direction (N/A because that means the request handler has nothing left to do) or finished on one direction and aborted in the other (e.g. read a fin, then get a STOP_SENDING - which you mention right below this).

Tests/NIOHTTPServerTests/ClientClosedTests.swift
```diff
@@ -0,0 +1,416 @@
+//===----------------------------------------------------------------------===//
```
ehaydenr, line 1: Can we add a test that actually goes through HTTPServerRequestHandler too?

0xTim: Just a heads up - we have a test for this in Vapor and it crashes around every 1 in 20 test runs with

```
NIOCore/NIOAsyncWriter.swift:177: Fatal error: Deinited NIOAsyncWriter without calling finish()
```

It only happens when under load (when we run the whole suit in parallel). The use case is the server handler sleeps for a second before responding, the client gives up after 200ms and closes the connection while the handler is still waiting, and the process traps shortly after the server logs a CancellationError for that request. Was planning on opening an actual issue once I had a reproducer if that helps

adam-fowler:
It seems like a bit of overkill allocating a new AsyncStream and additional child tasks for every connection given this is quite a niche situation ie the server has a connection open but is not reading from or writing to that connection. The majority of connections will never need this.

The only situations I've come across which needed this were where the request had been read and write hadn't started because either the server had to do a large amount of work before sending a response or it was serving SSE events which were fairly intermittent.

Is it worth making `withCancellationWhenClientCloses` a public function and let users call it when they actually need it so at least the additional child tasks are only allocated when needed. Also its signature could be changed to 

```swift
public func withClientCloseHandler<T>(
    operation: () async throws -> T,
    onClientClose handler: @Sendable @escaping () -> Void
) async rethrows -> T
```

Sources/NIOHTTPServer/NIOHTTPServer.swift
FranzBusch: Yeah that's fine with me

Sources/NIOHTTPServer/ClientClosed.swift
FranzBusch: I think this could be an argument to make [TaskCancellationScope](https://github.com/swiftlang/swift/blob/483fe2f2cda877d6a8599761c07342f7f5e94564/stdlib/public/Concurrency/TaskCancellationScope.swift#L31) public. In theory the only thing we need is a scope to cancel. Then we could give that scope to the underlying handler to cancel. This would require a slightly different setup potentially but I think it could work. In the end, for HTTP/1 it is something like this simplified:

```swift
let connection = try await server.nextHTTP1Connection()

try await withTaskCancellationScope { scope in
  try await withTaskGroup { group in
    group.addTask {
      try await connection.waitForClose()
      scope.cancel()
    }

    while true {
      let request = try await connection.nextRequest()
      try await requestHandler.handle(request, connection.reader, connection.writer)
    }
}
```

This way for an HTTP/1 connection we would only every have to spawn 1 child task that watches the connection closure and then just cancel the scope of the task that handles this connection.

@ktoso Would this be something that could motivate making `TaskCancellationScope`s public API?

Tests/NIOHTTPServerTests/ClientClosedTests.swift
FranzBusch: Yeah this is a bug somewhere in the HTTPServer then. We should fix it.

FranzBusch:
> ```swift
> public func withClientCloseHandler<T>(
>     operation: () async throws -> T,
>     onClientClose handler: @Sendable @escaping () -> Void
> ) async rethrows -> T
> ```

I generally like the idea of that. It would also compose with task cancellation scopes if we ever had them e.g.

```
func handle(request,...) async {
  await withTaskCancellationScope { scope in
    await withClientCloseHandler {
    } onClose {
      scope.cancel()
    }
  }
}
```

Sources/NIOHTTPServer/NIOHTTPServer.swift
gjcairo: This should have been "the _connection_ closing".

Tests/NIOHTTPServerTests/ClientClosedTests.swift
```diff
@@ -0,0 +1,413 @@
...
+import BasicContainers
+import Logging
+import NIOConcurrencyHelpers
+import NIOCore
+import NIOHTTP1
+import NIOHTTPTypes
+import NIOPosix
+import Testing
+
+@testable import NIOHTTPServer
```
FranzBusch, line 24: One of our goals was to not use `@testable` imports. Why do we need that?

FranzBusch (approved):
I think this looks generally good though except one nit around `@testable` usage. I think the behavior is the right one but the performance is something we have to work on. Thought I expect we have many more low hanging fruits than these child tasks

aryan-25 (approved):

Tests/NIOHTTPServerTests/ClientClosedTests.swift
```diff
@@ -0,0 +1,413 @@
...
+        #expect(
+            sawCancellation == false,
+            "STOP_SENDING is non-terminal and repeatable, so on its own it must not cancel the handler."
+        )
+    }
+
+    /// `STOP_SENDING` *after* the request end does cancel: with the client's FIN already received, the
+    /// exchange is finished in both directions, so the QUIC layer closes the stream — the `hasReceivedFin`
+    /// branch of `QUICChannelStreamHandler.receiveStopSending`, the one case where opting into half-closure
+    /// still closes.
```
aryan-25, line 193:
```suggestion
    /// `STOP_SENDING` *after* the request end does cancel: with the client's FIN already received, the
    /// exchange is finished in both directions, so the QUIC layer closes the stream.
```

Tests/NIOHTTPServerTests/ClientClosedTests.swift
gjcairo: For reference: https://github.com/swift-server/swift-http-server/issues/137

Thanks Tim

Tests/NIOHTTPServerTests/ClientClosedTests.swift
gjcairo: We don't, actually (:

ehaydenr (approved):

gjcairo:
@adam-fowler Thanks for bringing that up. I think it's a valid concern, and one we should look into. I believe with cancellation scopes, like Franz outlined above, this could become cheaper though, so I'd like to defer it for a while. Feel free to open an issue though and we can discuss it more there.

PR #135, opened 2026-09-14
Extract peerCertificateChain synchronously

### LINKED ISSUE #97: Change `peerCertificateChain` to be synchronous
From https://github.com/swift-server/swift-http-server/pull/92: we should change the `peerCertificateChain` in the `PeerCertificate` capability to not be async. We should be able to get the chain after the connection channel is set up.

aryan-25:
### Motivation
The peerCertificateChain property on `ConnectionContext`/`RequestContext` is marked as `async`. However, this is not necessary, because it is possible to synchronously obtain the peer's validated certificate chain from the connection channel.

### Modifications

- Updated the completion callback passed to the ALPN handler to also return the peer's certificate chain.
- Removed `async` and `throws` from the `peerCertificateChain` property in `ConnectionContext`, `RequestContext`, and the `HTTPServerCapability.PeerCertificate` protocol.

### Result
Fixes #97.


Sources/NIOHTTPServer/NIOHTTPServer+ConnectionContext.swift
```diff
@@ -53,31 +53,21 @@ extension NIOHTTPServer {
         /// The local address the connection is bound to, when known.
         public let localAddress: NIOHTTPServer.SocketAddress?
 
-        var peerCertificateChainFuture: EventLoopFuture<NIOSSL.ValidatedCertificateChain?>?
+        /// The peer's validated certificate chain. Returns `nil` if a custom verification callback was not set when
+        /// configuring mTLS in the server configuration, or if the custom verification callback did not return the
+        /// derived validated chain.
+        public var peerCertificateChain: X509.ValidatedCertificateChain?
```
FranzBusch: Somewhat orthogonal question but do we support extracting an unvalidated chain? There are use-cases where only certain routes in an HTTP server want authentication. If we require certificate chain validation for everything then that becomes impossible to support. So what if we had:
```swift
var validatedPeerCertificateChain: ...
var unvalidatedPeerCertificateChain: ...
```

aryan-25: > do we support extracting an unvalidated chain

Note that `peerCertificateChain` will only return a non-`nil` output if (1) a [custom certificate validation callback](https://github.com/swift-server/swift-http-server/blob/aa39e5c6c1934a2a06b26d2977c67bbc3846ab8b/Sources/NIOHTTPServer/Configuration/TransportSecurity%2BMTLSTrustConfiguration.swift#L117) was set and (2) a [successful result](https://github.com/swift-server/swift-http-server/blob/aa39e5c6c1934a2a06b26d2977c67bbc3846ab8b/Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift#L479) was returned from the callback.

This callback gets an array of all certificates presented by the peer, and the user has full control of the validation.

We also support an [optional verification mode](https://github.com/swift-server/swift-http-server/blob/aa39e5c6c1934a2a06b26d2977c67bbc3846ab8b/Sources/NIOHTTPServer/Configuration/NIOHTTPServerConfiguration.swift#L500-L502). In this mode, if the peer doesn't present certificates, they pass validation, but if they do present certificates, they must be valid. If you use the custom certificate callback alongside, you can also define what "valid" entails.

So the use-case you highlighted can indeed be supported by using the custom certificate callback and/or the optional verification mode.

FranzBusch: Okay then I think my remaining concern is about the property name or the documentation. It is called `peerCertificateChain` but it appears to be only the validated one. Should we rename this to `validatedPeerCertificateChain` then?

I can imagine use-cases which then just do very basic validation in the callback and do further validation in their request handlers.

aryan-25: I think the "validated" part should be quite clear. The name of the type is `ValidatedCertificateChain`.

That type doesn't have a "nice" public initializer. It just has [this](https://github.com/apple/swift-certificates/blob/c8aece90ea05f9866bd392a5bf13b5cae56c0e03/Sources/X509/Verifier/ValidatedCertificateChain.swift#L37-L50), which makes it very clear that the certificates passed to it must form a valid chain.

This is also the type returned by `swift-certificates`' `Verifier` upon a successful result.

I'm fine with renaming to `validatedPeerCertificateChain` if you really feel, but I think the name is slightly long and the "validated" part shouldn't reveal anything new.

FranzBusch: To me personally adding `validated` adds a lot of value here. Otherwise LGTM.

aryan-25: Fixed in commit [06d6c7c](https://github.com/swift-server/swift-http-server/pull/135/commits/06d6c7cf493f6c1c8af5ad1ced5ee94066752732).

FranzBusch (approved):

PR #136, opened 2026-09-18
Add concrete implementation of `sendAndFinish`

0xTim:
This adds a concrete implementation of `sendAndFinish` to avoid relying on the default protocol implementation which writes the head, body and trailers separately.

The PR combines all of those into a single write call instead of three separate writes that need to be written and flushed separately. In Vapor this produced a 15-22% increase in throughput with zero code changes on Vapor's side.


FranzBusch:
> The PR combines all of those into a single write call instead of three separate writes that need to be written and flushed separately. In Vapor this produced a 15-22% increase in throughput with zero code changes on Vapor's side.

Love to see this and exactly why we added the option to specialize this implementation

Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
```diff
@@ -66,6 +66,17 @@ extension NIOHTTPServer {
...
+
+        public func sendAndFinish<Buffer>(_ response: HTTPResponse, buffer: inout Buffer, trailer: HTTPFields?) async throws where Buffer : RangeReplaceableContainer, Buffer.Element == UInt8, Buffer : ~Copyable {
+            precondition(response.status.kind != .informational)
+            if buffer.isEmpty {
+                try await self.writer.write(contentsOf: [.head(response), .end(trailer)])
+            } else {
+                let body = ByteBuffer(draining: &buffer)
+                try await self.writer.write(contentsOf: [.head(response), .body(body), .end(trailer)])
+            }
+            self.writerState.wrapped.withLock { $0.finishedWriting = true }
```
FranzBusch, line 82: What happens if the above threw. Do we have to handle that?

Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
```diff
@@ -66,6 +66,17 @@ extension NIOHTTPServer {
...
             #endif
         }
+
+        public func sendAndFinish<Buffer>(_ response: HTTPResponse, buffer: inout Buffer, trailer: HTTPFields?) async throws where Buffer : RangeReplaceableContainer, Buffer.Element == UInt8, Buffer : ~Copyable {
+            precondition(response.status.kind != .informational)
+            if buffer.isEmpty {
+                try await self.writer.write(contentsOf: [.head(response), .end(trailer)])
+            } else {
+                let body = ByteBuffer(draining: &buffer)
+                try await self.writer.write(contentsOf: [.head(response), .body(body), .end(trailer)])
```
FranzBusch, line 80: There is even more room in the future here to optimize this since we should be able to get rid of this intermediate array allocation for the three parts

FranzBusch (approved):

FranzBusch (changes requested):
Can you fix up the formatting?

adam-fowler:
While I've been looking at the HTTP API proposal I wrote pretty much exactly the same piece of code.

Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
0xTim: If the write throws, `finishedWriting` stays false and the error propagates up to the caller. If it escapes the request handler, `invokeHandler` checks that flag and aborts so we are good. This also matches the existing `Writer.finish` implementation, which also only sets the flag after the write succeeds

Sources/NIOHTTPServer/NIOHTTPServerResponseSender.swift
0xTim: We can tweak this and get rid of an allocation but I think we'd need changes in NIO to get it down further

